PR #1, opened 2020-03-20
Typo and missing import in examples

dennistreysa:
[Fix] Typo
[Add] Missing import


lib/logic/tools/crypto/trithemius.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:gc_wizard/logic/tools/crypto/vigenere.dart';
+
+String encryptTrithemius(String input, {int aValue = 0}) {
+    return vigenereEncrypt(input, "ABCDEFGHIJKLMNOPQRSTUVWXYZ", false, aValue: aValue);
+}
```
S-Man42: For the  intents, we should follow the Dart convertions, don't you think?

lib/logic/tools/crypto/trithemius.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:gc_wizard/logic/tools/crypto/vigenere.dart';
+
+String encryptTrithemius(String input, {int aValue = 0}) {
+    return vigenereEncrypt(input, "ABCDEFGHIJKLMNOPQRSTUVWXYZ", false, aValue: aValue);
```
S-Man42: For standard alphabet: I tried to avoid such strings. I created some alphabet maps (`alphabet_AZ` in this case) which should be reused in all of the cases. Maybe an optional parameter makes sense here for normal cases.

Maybe you could do something like:
`var key = alphabet_AZ.entries.map((entry) => entry.key).join();`

lib/widgets/registry.dart
```diff
@@ -115,6 +116,7 @@ class Registry {
       GCWToolWidget(tool: ESelection(), toolName: i18n(context, 'e_selection_title'), searchStrings: SEARCHSTRING_E),
       GCWToolWidget(tool: Enigma(), toolName: i18n(context, 'enigma_title'), searchStrings: 'enigma rotors walzen'),
       GCWToolWidget(tool: KennysCode(), toolName: i18n(context, 'kennyscode_title'), searchStrings: 'they killed kenny sie haben kenny getötet kennys kenny\'s code southpark'),
+      GCWToolWidget(tool: Trithemius(), toolName: i18n(context, 'trithemius_title'), searchStrings: SEACHSTRING_ROTATION + 'trithemius'),
```
S-Man42: I would add 'vigenere' here as well... otherwise it wouldn't be shown when search for v.

test/logic/tools/crypto/trithemius.dart
```diff
@@ -0,0 +1,80 @@
+import "package:flutter_test/flutter_test.dart";
+import 'package:gc_wizard/logic/tools/crypto/trithemius.dart';
+
+void main() {
+  group("Trithemius.encrypt:", () {
+    List<Map<String, dynamic>> _inputsToExpected = [
+      {'input' : null, 'aValue': 0, 'expectedOutput' : ''},
```
S-Man42: maybe add a check for `aValue == null` or `aValue < 0`?

dennistreysa: Is it possible that and int becomes null?
I guess aValue < 0 is valid.

lib/widgets/registry.dart
dennistreysa: Mh, in case the user is looking for something that "looks like Vigenere, but quite isn't" it makes sense to add it. So yes... good idea.

lib/logic/tools/crypto/trithemius.dart
dennistreysa: Yes, was think about something like this, too. 👍

lib/logic/tools/crypto/trithemius.dart
dennistreysa: Yes, sure. If the Dart convention is the intended project indention style, we can stick to that. Was probably a 'typo' :D

PR #2, opened 2020-03-21
[add] Trithemius Cipher

dennistreysa:
Trithemius is a simple variant of the Vigenere Cipher, thus the original Vigenere methods were used.

### DIFF
--- a/lib/logic/tools/crypto/trithemius.dart
+++ b/lib/logic/tools/crypto/trithemius.dart
@@ -1,0 +1,9 @@
+import 'package:gc_wizard/logic/tools/crypto/vigenere.dart';
+
+String encryptTrithemius(String input, {int aValue = 0}) {
+    return vigenereEncrypt(input, "ABCDEFGHIJKLMNOPQRSTUVWXYZ", false, aValue: aValue);
+}
+
+String decryptTrithemius(String input, {int aValue: 0}) {
+    return vigenereDecrypt(input, "ABCDEFGHIJKLMNOPQRSTUVWXYZ", false, aValue: aValue);
+}
--- a/lib/logic/tools/crypto/trithemius.dart
+++ b/lib/logic/tools/crypto/trithemius.dart
@@ -1,9 +1,12 @@
 import 'package:gc_wizard/logic/tools/crypto/vigenere.dart';
+import 'package:gc_wizard/utils/alphabets.dart';
+
+final trithemiusKey = alphabet_AZ.entries.map((entry) => entry.key).join();
 
 String encryptTrithemius(String input, {int aValue = 0}) {
-    return vigenereEncrypt(input, "ABCDEFGHIJKLMNOPQRSTUVWXYZ", false, aValue: aValue);
+  return vigenereEncrypt(input, trithemiusKey, false, aValue: aValue);
 }
 
 String decryptTrithemius(String input, {int aValue: 0}) {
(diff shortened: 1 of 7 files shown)


lib/logic/tools/crypto/trithemius.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:gc_wizard/logic/tools/crypto/vigenere.dart';
+
+String encryptTrithemius(String input, {int aValue = 0}) {
+    return vigenereEncrypt(input, "ABCDEFGHIJKLMNOPQRSTUVWXYZ", false, aValue: aValue);
+}
```
S-Man42: 1. Intents: I would stay with the Dart conventions and the two spaces... Well, I know, it's ugly...
1. For the standard alphabet, I tried to avoid, retype this string over and over again. Theres is a map, which maps the letters to their letter values. This one is used many times. In this case you could do:
`var key = alphabet.entries.map((entry) => entry.key).join();`

lib/widgets/registry.dart
```diff
@@ -115,6 +116,7 @@ class Registry {
       GCWToolWidget(tool: ESelection(), toolName: i18n(context, 'e_selection_title'), searchStrings: SEARCHSTRING_E),
       GCWToolWidget(tool: Enigma(), toolName: i18n(context, 'enigma_title'), searchStrings: 'enigma rotors walzen'),
       GCWToolWidget(tool: KennysCode(), toolName: i18n(context, 'kennyscode_title'), searchStrings: 'they killed kenny sie haben kenny getötet kennys kenny\'s code southpark'),
+      GCWToolWidget(tool: Trithemius(), toolName: i18n(context, 'trithemius_title'), searchStrings: SEACHSTRING_ROTATION + 'trithemius'),
```
S-Man42: I would add *vigenere* here as well... So, you'll get this result, when you're search von v. Don't you think?

test/logic/tools/crypto/trithemius.dart
```diff
@@ -0,0 +1,80 @@
+import "package:flutter_test/flutter_test.dart";
+import 'package:gc_wizard/logic/tools/crypto/trithemius.dart';
+
+void main() {
+  group("Trithemius.encrypt:", () {
+    List<Map<String, dynamic>> _inputsToExpected = [
+      {'input' : null, 'aValue': 0, 'expectedOutput' : ''},
```
S-Man42: Maybe one additional test case for `aValue == null` and one for `aValue < 0`?

dennistreysa: Is it possible that and int becomes null?
I guess aValue < 0 is valid (due to the power of modulo: -1 === 25)

lib/widgets/registry.dart
dennistreysa: Mh, in case the user is looking for something that "looks like Vigenere, but quite isn't" it makes sense to add it. So yes... good idea.

test/logic/tools/crypto/trithemius.dart
S-Man42: Yes, in Dart `int` is an object like everything else and so can become `null` theoretically.

lib/logic/tools/crypto/trithemius.dart
S-Man42: Hab grad überlegt, genau diese Codezeile nochmal auszulagern, weil man die öfter braucht... Mach ich mal beizeiten

PR #3, opened 2020-03-21
[fix] Typo in ADFGVX

dennistreysa:
A comparison instead of an assignment was used. This goes unnoticed by the unit tests since the createPolybiosAlphabet method  handles it.

### DIFF
--- a/lib/logic/tools/crypto/adfgvx.dart
+++ b/lib/logic/tools/crypto/adfgvx.dart
@@ -38,7 +38,7 @@ String _encrypt (
     return '';
 
   if (substitutionKey == null)
-    substitutionKey == '';
+    substitutionKey = '';
 
   var adfgvxMode = mode.toString().split('.')[1]; //mode.toString() == _ADFGVX.ADFGX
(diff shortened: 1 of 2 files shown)


lib/logic/tools/crypto/adfgvx.dart
```diff
@@ -38,7 +38,7 @@ String _encrypt (
     return '';
 
   if (substitutionKey == null)
-    substitutionKey == '';
+    substitutionKey = '';
```
S-Man42: Great catch! Now, we now, the unit test was not well enough. We should enhance it, as well to avoid that error in future refactorings.

dennistreysa: Couldn't think of a way to do it, but yes, better tests is always better ;)

S-Man42: I'll try one. :)

S-Man42: Yeah, impossible to catch explicitly at this point. But the called method in Polybios cipher can be tested instead. Commited these tests. Are you ok with this?

dennistreysa: I guess it's sufficient enough ;)

PR #4, opened 2020-03-22
Gronsfeld Cipher

dennistreysa:
I added the Gronsfeld Cipher which is basically Vigenere with digits (0..9 = A..J).
Also a function to convert a string with digits to a string containing the alpha equivalents was added.


lib/utils/common_utils.dart
```diff
@@ -25,6 +25,31 @@ String intListToString(List<int> list, {String delimiter: ''}) {
   return list.map((elem) => elem == null ? unknownElement : elem).join(delimiter).trim();
 }
 
+String digitsToAlpha(String text, {int aValue: 0, bool removeNonDigits: true}) {
```
S-Man42: I would add a test for this (and maybe for all missing util methods as well :D)

lib/widgets/tools/crypto/gronsfeld.dart
```diff
@@ -0,0 +1,104 @@
...
+      children: <Widget>[
+        GCWTextField(
+          controller: _inputController,
+          onChanged: (text) {
+            setState(() {
+              _currentInput = text;
+            });
+          },
+        ),
+        GCWTextField(
```
S-Man42: You could use GCWIntegerTextField, couldn't you? And then: _currentKey = value['text']

lib/utils/common_utils.dart
dennistreysa: I added a test.

lib/widgets/tools/crypto/gronsfeld.dart
dennistreysa: Sure. I just hadn't figured out that there is a "base" folder and all the derived classed are outside of that. I changed it to GCWIntegerTextField which is indeed more convenient.

PR #7, opened 2020-03-22
Lemon cipher

dennistreysa:
The Lemon Cipher is somewhat unknown. It originates from a letter sent to Franklin D. Roosevelt: http://scienceblogs.de/klausis-krypto-kolumne/2015/06/22/geloest-der-verschluesselte-drohbrief-an-us-praesident-roosevelt/

I don't know the proper name, but one day I got introduced to it as "Lemon Cipher".
The Clockwise/Counterclockwise option was an idea by myself.

There exists a tool to encrypt/decrypt: https://www.kryptographiespielplatz.de/index.php?aG=004a42fdaff92112591b4cac0e611ab0a648fe27

### DIFF
--- a/assets/i18n/de.json
+++ b/assets/i18n/de.json
@@ -393,5 +393,10 @@
   "enigma_numberrotors" : "Anzahl",
   "enigma_input" : "Eingabe",
 
-  "trithemius_title" : "Trithemius Chiffre"
+  "trithemius_title" : "Trithemius Chiffre",
+
+  "lemon_title" : "Lemon Chiffre",
+  "lemon_mode" : "Richtung",
+  "lemon_clockwise" : "UZS",
+  "lemon_counter_clockwise" : "geg. UZS"
 }
--- a/lib/logic/tools/crypto/lemon.dart
+++ b/lib/logic/tools/crypto/lemon.dart
@@ -1,0 +1,57 @@
+import 'dart:math' as math;
+import 'package:gc_wizard/logic/tools/crypto/rotator.dart';
+import 'package:gc_wizard/utils/common_utils.dart';
+
+String encryptLemon(String input, {bool counterClockWise: true}) {
+  if ((input == null) || (counterClockWise == null))
+    return '';
+
+  input = removeNonAlphaNum(input);
+  
+  if (input.length <= 1)
+    return input;
+  
+  String odd;
+  String even;
+  
+  if (counterClockWise) {
+    final int half = (input.length / 2).floor();
+    even = input.substring(0, half);
+    odd = input.substring(half).split('').reversed.join('');
+  } else {
+    final int half = (input.length / 2).ceil();
+    odd = input.substring(0, half);
+    even = input.substring(half).split('').reversed.join('');
+  }
+  
+  String zipped = '';
+  
+  for ( int index = 0; index < math.max(odd.length, even.length); index++) {
+    zipped += (even.length > index) ? even[index] : '';
+    zipped += (odd.length > index) ? odd[index] : '';
+  }
+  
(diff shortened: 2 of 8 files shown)


dennistreysa:
I just realized the tool from Marcel Brätz works different then my tool for strings with odd length.

So for example `ABCDE` encrypts to `AEBDC`.
Which is wrong in my opinion.

I gues he initially does this like:

```
AE
BD
C
```

Which.... now that I look at it.... is exactly what I do in the "counterclockwise" mode.

I guess thats the reason I implemented both versions in the Multisolver back in 2015 ;)

lib/logic/tools/crypto/lemon.dart
```diff
@@ -0,0 +1,57 @@
+import 'dart:math' as math;
+import 'package:gc_wizard/logic/tools/crypto/rotator.dart';
+import 'package:gc_wizard/utils/common_utils.dart';
+
+String encryptLemon(String input, {bool counterClockWise: true}) {
```
S-Man42: I would reuse a more global function, like the skytale function (maybe extract the skytale as global transposition function):

```
encrpytLemon{
   input = secondHalfOfInput.reverse + firstHalf
   return encryptSkytale(input, columnCount: input.length, countLettersPerCell: 1)
}
```

Just define what to do with odd number of input length.

assets/i18n/de.json
```diff
@@ -393,5 +393,10 @@
   "enigma_numberrotors" : "Anzahl",
   "enigma_input" : "Eingabe",
 
-  "trithemius_title" : "Trithemius Chiffre"
+  "trithemius_title" : "Trithemius Chiffre",
+
+  "lemon_title" : "Lemon Chiffre",
+  "lemon_mode" : "Richtung",
+  "lemon_clockwise" : "UZS",
```
S-Man42: Is "UZS" a common term? If you hadn't read "CW" before, I am not sure if I knew what that should mean :D

lib/widgets/registry.dart
```diff
@@ -118,6 +119,7 @@ class Registry {
       GCWToolWidget(tool: Enigma(), toolName: i18n(context, 'enigma_title'), searchStrings: 'enigma rotors walzen'),
       GCWToolWidget(tool: KennysCode(), toolName: i18n(context, 'kennyscode_title'), searchStrings: 'they killed kenny sie haben kenny getötet kennys kenny\'s code southpark'),
       GCWToolWidget(tool: Trithemius(), toolName: i18n(context, 'trithemius_title'), searchStrings: SEARCHSTRING_VIGENERE + 'trithemius tabula recta'),
+      GCWToolWidget(tool: Lemon(), toolName: i18n(context, 'lemon_title'), searchStrings: 'lemon franklin roosevelt'),
```
S-Man42: maybe +transposition?

lib/widgets/tools/crypto/lemon.dart
```diff
@@ -0,0 +1,80 @@
...
+        _buildOutput(context)
+      ],
+    );
+  }
+
+  Widget _buildOutput(BuildContext context) {
+    var output = '';
+
+    if (_currentMode == GCWSwitchPosition.left) {
+      output = encryptLemon(_currentInput, counterClockWise: (_direction != GCWSwitchPosition.left));
```
S-Man42: I would prefer positive statements whenever possible: _direction == GCWSwitchPosition.right. Is more readable, I believe

S-Man42:
Close this because of conflicting state and unsolved questions.

Would love to see this in the GCW, nonetheless. Please feel free to reopen this again if you like!

PR #24, opened 2020-03-24
Symbol Tables

S-Man42:
Symbol tables and some minor refactorings


lib/widgets/tools/crypto/symbol_table.dart
```diff
@@ -0,0 +1,263 @@
...
+
+    for (var i = 0; i <= countRows; i++) {
+      var columns = <Widget>[];
+
+      for (var j = 0; j < countColumns; j++) {
+        var widget;
+        var imageIndex = i * countColumns + j;
+
+        if (imageIndex < _currentInput.length) {
+          var imagePath = _imageFilePaths[_currentInput.toUpperCase().codeUnitAt(imageIndex).toString()];
```
dennistreysa: Maybe pull `_currentInput.toUpperCase()` out to not calculate it every round?
Also, is there a reason why the key needs to be String? Maybe it can be minimal-optimized if it was int?

dennistreysa:
So far looking very impressive!

You think it is possible to have a combined input? Symbol-Buttons (like in Decrypt), but with Letters displayed over/next to them? This could speed up typing if you already know the last few letters of a word.

lib/widgets/tools/crypto/symbol_table.dart
S-Man42: Second first: At least it is necessary to accept strings. It is possible to get symbols, that a not really letters or something like that. In semaphore, for example, there are flags for *ERROR*, *CORRECT*, *CANCEL*. So there are possible keys: `65`, `32`, but also `correct`

S-Man42: First: I cannot do something like _currentInput = _currentInput.toUpperCase(), because this would lead to inconsistent states. This variable shows the real input of the user.

I added a new local variable which keeps the uppercase instead...

dennistreysa: That's what I meant ;)

lib/widgets/tools/crypto/symbol_table.dart
```diff
@@ -183,11 +198,18 @@ class SymbolTableState extends State<SymbolTable> {
     );
   }
 
-  _getSpecialText(key) {
-    return i18n(context, 'symboltables_' + widget.symbolKey + '_' + key);
+  _getSpecialText(key, {showDelimiterOnSpecialText: true}) {
+    var specialText = i18n(context, 'symboltables_' + widget.symbolKey + '_' + key);
```
dennistreysa: This returns `null` when the string could not be found and breaks the code after.
Maybe add an `specialText ??= key;` ?

S-Man42: Yes, this breaks everytime a key is not found in the localization function. For me it's great. Because it MUSTN'T break. This shows an missing key and has to be fixed... Therefore, I don't want to fix it ;)

Issue #6, opened 2020-03-22
textToIntList needs reworking.


dennistreysa:
`textToIntList` in common_utils.dart lookes somewhat complicated to me. Instead of using a replace approach, I would suggest an extraction approach by the following code:

```dart
List<int> textToIntList(String text, {bool allowNegativeValues: false}) {
  if ((text == null) || (allowNegativeValues == null))
    return [];
      
  final regex = allowNegativeValues ? RegExp(r'-?\d+') : RegExp(r'\d+');
  
  return regex.allMatches(text).map((value) => int.tryParse(text.substring(value.start, value.end))).toList();
}
```

That also eliminates the edge case of a straying '-'.

S-Man42:
Hm, to be honest: For me, this approach looks just as complicated to me as mine or vice versa... I see neither advantages for yours nor for mine. So, it's not clear to me, why you would change this. Can you explain it? :)

dennistreysa:
I find it more convenient to have a RegEx that is "extracting what I'm looking for" and not "throwing away what I'm NOT looking for", especially when the method implies that by it's naming.
The opposite may apply to special "cleaner"-methods like "removeNonAlpha" etc.

Regarding your objection, that this will break the user being able to insert negative values:

The `GCWIntegerListTextField` currently does not remove any chars on it's own, so no worries, but I'm not sure if this really was the intent.

I think the real problem is the interaction between `textToIntList` `GCWIntegerListTextField` and `IntegerTextInputFormatter`.

Which makes this Issue related to #5 

So, my idea here is the following:

`textToIntList` shall extract only valid numbers (like in my code above)
`IntegerTextInputFormatter` shall accept anything that could become a valid number (similar to your 
textToIntList) and filter out everything else.

Which in return means "1 - -2 -   3" is a valid input, but shall be parsed to [1, -2, 3] or [1, -2, -3] (I leave that to you, but would go with second).

Which leads me to another problem. What if the user copies and pastes a list that looks like the following: "1, 2, 3, 4" (which looks like a valid, parsable integer list to me).

As far as I understood the code, the input formatter would fall back to the original string and one would not be able to paste such a string (because of the commas).

I guess the solution here is to return a new filtered string instead of just the old one.

S-Man42:
This should be discussed from beginning, since the development went on and any thoughts may be deprecated.

Please feel free to reopen it, if necessary

PR #25, opened 2020-03-25
Reworked IntegerTextInputFormatter

### LINKED ISSUE #5: IntegerTextInputFormatter needs reworking/explanation
I tried to understand what's going on in the `formatEditUpdate` method of the class `IntegerTextInputFormatter`. The 'normal' mode seems to work properly, but list mode is somewhat broken, or at least I don't quite get it.

The following piece

```dart
newValue.text.split(regex).forEach((value) {
  if (!_checkIntegerValue(newValue.text))
    return oldValue;
});
```

does nothing. I guess the intention here was to return the old value back to the caller of `formatEditUpdate` which is not the case. Also `newValue.text` should be replaced by `value`I guess.

What is the intent here?

dennistreysa:
I tried to solve common problems that arise from a life-formatting Integer(List)-Input.

This is what I came up with:

- IntegerTextInputFormatter does not check for range anymore, this is done by the parser in the onChanged-callback. (Maybe we can change this, but I have to think about it again :D )
- The formatter itself works like follows:
    - Replace everything that we don't need with space
    - Cleanup (double space etc.)
    - Fix the cursor position
- Strings like `"---"` and other variations of it are still valid, but textToIntList will handle this properly.

I guess this works a bit smoother than before. However, I don't say it's perfect and there are surely some strange edge-cases I haven thought through.

But for now, I think this resolves #5 and the ideas discussed in #6

### DIFF
--- a/lib/utils/common_utils.dart
+++ b/lib/utils/common_utils.dart
@@ -112,9 +112,11 @@ String insertSpaceEveryNthCharacter(String input, int n) {
   return out.trim();
 }
 
-String sanitizeIntegerString(String input, bool allowNegativeValues, bool allowNumberList) {
+String sanitizeIntegerString(String input, {bool allowNegativeValues: true, bool allowNumberList: true}) {
   
   String adaptedText = input ?? '';
+  allowNegativeValues ??= true;
+  allowNumberList ??= true;
   
   if (allowNegativeValues) {
     if (allowNumberList) {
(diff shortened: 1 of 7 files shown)


S-Man42:
This seems not working properly. 

Current behaviour is:
1. You cannot delete the numbers with backspace.
1. You can delete the input by selecting and deleting it, but then the value is set to minimum.

With your code, try following:

1. Go to "Settings -> General -> Fontsize" (its an integertextfield, min 10, max 30). 
1. If there's a value like "16": Try to set the cursor behind the 6 and then delete characters using backspace.
1. You are able to delete the 6, having a lonely 1. This mustn't be, because 10 is minimum. 
1. Furthermore, you are able to delete the 1 as well, which breaks the app.

dennistreysa:
> This seems not working properly.

There was an edge case with an empty input, should be fixed now. The solution is to first fall back to min, if that's not present, fall back to max and if that's not present, too, fall back to 0.

S-Man42:
There's a handlePaste() method, which might help https://api.flutter.dev/flutter/widgets/TextSelectionControls/handlePaste.html

S-Man42:
It needs to be a complete rivisiting... Closing for now but should be done one time

PR #26, opened 2020-03-25
TomTom Code (also known was A-Tom-Tom)

dennistreysa:
I added the TomTom code (also known as A-Tom-Tom). https://www.omniglot.com/conscripts/atomtom.htm

Additionally I added a helper function for swapping out lookup maps.


dennistreysa:
Since it was not much work to do, I added Abaddon right away ;)

PR #138, opened 2020-08-04
Update README.md

t-m-z:
Zwei Anpassungen:
- Hinweis auf Installation API Level 28
- Registrierung des Widgets -> Ergänzung der import-Anweisung


S-Man42:
Thanks for that. Installing the SDK should be included in the very first step (following instructions). I made it clearer at this point. So, I close this PR, but added your advices to the README nevertheless.

PR #144, opened 2020-08-06
Segment

t-m-z:
Provide a new feature: encoding and decoding of 7-, 14- and 16-Segment-Displays

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,6 @@
 web
 /pubspec.lock
 /assets/tokens/mapbox
+*.tmp
+*.docx
+*.pdf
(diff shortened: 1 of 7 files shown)


t-m-z:
the app is running; however some issues have still to be discussed:

- is there a common understanding how digits have to be encoded? Currently I rely on the encoding from http://kryptografie.de/kryptografie/chiffre/14-segment.htm

- how flexible should be the decoding and who provides the definitions?

- should we stick to lower case?

S-Man42:
I merge this and afterwards I'll merge it with my UI branch. Let's see how the result looks like :D

1. I've no idea :D
2. As much as possible. I believe we have to add some decode codes from time to time... Let's start with yours
3. Yes, I'd do so. I seems the common case.

PR #145, opened 2020-08-07
Segment display

t-m-z:
finish   final Map<String, List<String>> AZTo7Segment 

Source: https://www.wikizero.com/en/Seven-segment_display


lib/logic/tools/science_and_technology/segment_display.dart
```diff
@@ -26,13 +26,55 @@ final Map<String, List<String>> AZTo14Segment = {
...
 final Map<String, List<String>> AZTo7Segment = {
-  '1' : ['b','c'], '2' : ['a','b','d','e','g'], '3' : ['a','b','c','d','g'], '4' : ['b','c','f','g'], '5' : ['a','c','d','f','g'],
-  '6' : ['a','c','d','e','f','g'], '7' : ['a','b','c'], '8' : ['a','b','c','d','e','f','g'], '9' : ['a','b','c','d','f','g'], '0' : ['a','b','c','d','e','f'],
-  'A' : ['a','b','c','e','f','g'], 'B' : ['c','d','e','f','g'], 'C' : ['a','d','e','f'], 'D' : ['b','c','d','e','g'], 'E' : ['a','d','e','f','g'],
-  'F' : ['a','e','f','g'], 'G' : ['a','c','d','e','f'], 'H' : ['b','c','e','f','g'], 'I' : ['a','e'], 'J' : ['a','c','d'], 'K' : ['a','c','e','f','g'],
-  'L' : ['d','e','f'], 'M' : ['a','c','e','g'], 'N' : ['a','b','c','e','f'], 'O' : ['c','d','e','g'], 'P' : ['a','b','e','f','g'], 'Q' : ['a','b','c','f','g'],
-  'R' : ['a','g'], 'S' : ['a','c','d','f'], 'T' : ['d','e','f','g'], 'U' : ['b','c','d','e','f'], 'V' : ['b','e','f','g'], 'W' : ['b','d','f','g'], 'X' : ['c','e'],
-  'Y' : ['b','c','d','f','g'], 'Z' : ['a','b','d','e'], ' ' : []
+  // https://www.wikizero.com/en/Seven-segment_display
+  '1'  : ['b','c'], '1' : ['a','e'],
```
S-Man42, line 30: This won't work, keys needs to be unique. Otherwise you'll create a warning and undefined behaviour concerning the returned value

That's why the SegmentToAZ contains more elements than AZToSegment. The way back is correct.

Other way round you cannot define the same list for different characters in SegmentToAZ (e.g. for C, ( and [ )

So, in both ways we have to decide for one single value if there were more choices.

t-m-z: Okay, I thougt dart is much more powerful ragarding maps. However, this eases the discussion. 

So we have to build 6 maps:
3 maps AZTo Segment(7|14|16) where we set the standard
3 maps  Segment(7|14|16)ToAZ where we try to catch every possible representation of a digit, letter, symbol etc.
No Problem.

PR #146, opened 2020-08-07
display basic error messages

t-m-z:
if the users starts an ecoding or decoding without an input or with an empty custom alphabet, gcw displays an error

### DIFF
--- a/lib/widgets/tools/crypto_and_encodings/bifid.dart
+++ b/lib/widgets/tools/crypto_and_encodings/bifid.dart
@@ -136,8 +136,10 @@ class BifidState extends State<Bifid> {
 
         GCWEncryptButtonBar(
           onPressedEncode: () {
-            if (_currentInput == null || _currentInput.output.length == 0) {
-              _currentOutput = BifidOutput("bifid_error_no_encrypt_input", null);
+            if (_currentInput == null || _currentInput.length == 0) {
+              setState(() {
+                _currentOutput = errorBifid("bifid_error_no_encrypt_input");
+              });
             } else {
               if (_currentMatrixMode == GCWSwitchPosition.left) {
                 _currentKey = "12345";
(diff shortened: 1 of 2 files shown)


lib/widgets/tools/crypto_and_encodings/bifid.dart
```diff
@@ -174,25 +205,32 @@ class BifidState extends State<Bifid> {
         text: '' //TODO: Exception
       );
     } else {
+      // nutzung von Toasts - https://github.com/t-m-z/GCWizard/blob/_Bifid/lib/widgets/tools/formula_solver/formula_solver_formulagroups.dart
+      // Zeile 122ff
+
       switch (_currentOutput.output) {
```
S-Man42, line 211: Maybe an even better approach to avoid redundancy: (Pseudo code)

```
if (BifidOutput.state == STATE_OK) {
  return <valid output>;
} else {
  showToast(i18n(context, BifidOutput.errorMessage));
  return GCWDefaultOutputput(text: ''); 
}
```

Of couse, a map based solution is also ok:

`bifidOutput['state']`

I did something similar already right here: https://github.com/S-Man42/GCWizard/blob/master/lib/logic/tools/formula_solver/parser.dart#L83

The advantage is 
a) You don't need the switch statement with nearly the same content
b) You can add error messages without need to change the UI code.

lib/widgets/tools/crypto_and_encodings/bifid.dart
```diff
@@ -136,8 +137,11 @@ class BifidState extends State<Bifid> {
 
         GCWEncryptButtonBar(
           onPressedEncode: () {
-            if (_currentInput == null || _currentInput.output.length == 0) {
-              _currentOutput = BifidOutput("bifid_error_no_encrypt_input", null);
+            // nutzung von Toasts - https://github.com/t-m-z/GCWizard/blob/_Bifid/lib/widgets/tools/formula_solver/formula_solver_formulagroups.dart
```
S-Man42, line 140: Please remove such comments. They seem to be just a reminder for you, but contain no further information concerning the code.

lib/widgets/tools/crypto_and_encodings/bifid.dart
t-m-z: you are right - thank you for this hint

PR #150, opened 2020-08-10
Implement a new Feature: BCD en-/decoding: Original, AIken, Stibitz, …

t-m-z:
I am not quite sure that I met all your standards so please feel free to comment.
Also I am still not sure, that I am running git properly. 
Finally two questions:

1. because you get a direct result from your input, are tests necessary? 

1. what's about Gray, 1-of-10, and other codes - should they be integrated in the BCD-Submenu or an stand-alone-menu item?


lib/logic/tools/crypto_and_encodings/bcd.dart
```diff
@@ -0,0 +1,138 @@
...
+       var character = BCDStibitzToDigit[bcd];
+
+        return character != null ? character : '';
+      })
+      .join();
+}
+
+
+String BCDdecodeLibawCraig(String input){
+  if (input == null || input == '')
```
S-Man42, line 123: Because the Encode/Decode codes are always the same, it could make sense to avoid replication like this:

```
enum _BCDType {ORIGINAL, AIKEN, ...}

_encodeBCD(String input, _BCDType type) {
  if (input == null || input == '')
    return '' ;

  var bcdMap;
  switch (type) {
    case _BCDType.AIKEN: bcdMap = DigitToBCDAiken;
   ...
  }

  return input
      .split('')
      .where((character) => bcdMap[character] != null)
      .map((character) => bcdMap[character])
      .join(' ');
}

String encodeBCDAiken(String input) {
   return _BCDencode(input, _BCDType.AIKEN);
}
```

lib/logic/tools/crypto_and_encodings/bcd.dart
```diff
@@ -0,0 +1,138 @@
...
+
+        var character = BCDAikenToDigit[bcd];
+
+        return character != null ? character : '';
+      })
+      .join();
+}
+
+
+String BCDdecodeStibitz(String input){
```
S-Man42, line 104: Please rename to `decodeBCDStibitz` (verb should be first an non-capital letters)

lib/widgets/registry.dart
```diff
@@ -136,6 +141,7 @@ class Registry {
   static final SEARCHSTRING_ASTRONOMY_RISESET = 'rise set transit noon aufgang aufgaenge untergang untergaenge dawn dusk mittag culmination kulmination ';
   static final SEARCHSTRING_ASTRONOMY_POSITION = 'declination position stand rektaszension zodiac astrology astrologischeszeichen astrologie astrologicalsign tierkreiszeichen sternzeichen azimuth altitude diameter durchmesser hoehe rightascension distance entfernung distanz observer beobachter juliandate julianischesdatum sternzeit siderealtime gmst lmst eclipticlongitude ekliptischelaenge ';
   static final SEARCHSTRING_BASE = 'base encode decode encoding decoding dekodierung dekodieren ';
+  static final SEARCHSTRING_BCD = 'bcd encode decode encoding decoding dekodierung dekodieren ';
```
S-Man42, line 144: maybe add something like binary coded decimal, binär, ...

lib/widgets/tools/crypto_and_encodings/bcd/bcdlibawcraig.dart
```diff
@@ -0,0 +1,67 @@
+import 'package:flutter/material.dart';
+import 'package:gc_wizard/logic/tools/crypto_and_encodings/bcd.dart';
```
S-Man42, line 2: Please try to reuse code also in frontend. You could create a BCD widget with a type parameter. At the end, theres only a switch for the correct logic call.

I did something similar e.g. with the segment display:

https://github.com/S-Man42/GCWizard/blob/1.1.0/lib/widgets/tools/science_and_technology/segment_display/seven_segments.dart

https://github.com/S-Man42/GCWizard/blob/1.1.0/lib/widgets/tools/science_and_technology/segment_display/segment_display.dart

assets/i18n/de.json
```diff
@@ -171,6 +171,16 @@
   "base_base85_title" : "Base85 (ASCII85)",
   "base85_example" : "<~<+oue+DGm>FD,5.@;^.1AT@~>",
 
+  "bcd_selection_title" : "BCD Codierungen",
+  "bcd_original_title" : "BCD",
```
S-Man42, line 175: may description: "Encodes numbers to several binary representations" or somehing like that.

lib/widgets/tools/crypto_and_encodings/bcd/bcdlibawcraig.dart
```diff
@@ -0,0 +1,67 @@
...
+  void dispose() {
+    _inputController.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      children: <Widget>[
+        GCWTextField(
```
S-Man42, line 34: Maybe it could be good idea to use two different textfields according to the encrypt state. One that allows [0-9] and one  which allows [01 ].

```
_currentMode == GCWSwitchPosition.left 
  ? GCWTextField(
       controller: _encodeController;
       inputFormatters: [_encodeMaskFormatter ]
     )
  : GCWTextField (
      controller: decodeController;
      inputFormatters: [_decodeMaskFormatter ]
    ),
```

For the inputformatters you could use the MaskTextInputFormatter like in
https://github.com/S-Man42/GCWizard/blob/1.1.0/lib/widgets/tools/crypto_and_encodings/gc_code.dart#L22 (used in line 46)

For encode:

```
  var _encodeMaskFormatter = MaskTextInputFormatter(
    mask: '#' * 10000, // allow 10000 characters input
    filter: {"#": RegExp(r'[0-9]')}
  );
```

for decode:

```
  var _decodeMaskFormatter = MaskTextInputFormatter(
    mask: '#### ' * 5000, // allow 5000 4-digit binary blocks, spaces will be set automatically after each block
    filter: {"#": RegExp(r'[01]')}
  );
```

S-Man42:
Furthermore, you always need to merge the current 1.1.0 branch into your feature branch before the PR to avoid conflicts. The 1.1.0 changes because of the other merged features. This may result in changes in the same file. In this project that's nearly everytime the registry.dart and the main_view.dart, because they are changed for every new feature. 

Currently there's a conflict that needs to be resolved before being able to merge this.

assets/i18n/de.json
```diff
@@ -171,6 +171,16 @@
   "base_base85_title" : "Base85 (ASCII85)",
   "base85_example" : "<~<+oue+DGm>FD,5.@;^.1AT@~>",
 
+  "bcd_selection_title" : "BCD Codierungen",
+  "bcd_original_title" : "BCD",
+  "bcd_original_example" : "",
+  "bcd_aiken_title" : "BCD Variante nach Aiken",
```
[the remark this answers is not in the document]
t-m-z: Done - also in en.json

lib/logic/tools/crypto_and_encodings/bcd.dart
```diff
@@ -0,0 +1,138 @@
+// https://de.wikipedia.org/wiki/BCD-Code
+// https://de.wikipedia.org/wiki/Aiken-Code
+// https://de.wikipedia.org/wiki/Stibitz-Code
+// https://de.wikipedia.org/wiki/Libaw-Craig-Code
+// https://de.wikipedia.org/wiki/Gray-Code
```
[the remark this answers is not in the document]
t-m-z: Done

lib/logic/tools/crypto_and_encodings/bcd.dart
```diff
@@ -0,0 +1,138 @@
...
+
+
+String BCDdecodeAiken(String input){
+  if (input == null || input == '')
+    return '';
+
+  return input
+      .split(RegExp(r'[^01]'))
+      .map((bcd) {
+        if (bcd == '.' || bcd == ' '  || bcd == '-' )
```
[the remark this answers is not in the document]
t-m-z: No - I missunderstood the RegExp. Thus the if-construct is not necessary

lib/logic/tools/crypto_and_encodings/bcd.dart
t-m-z: Great idea - done

lib/logic/tools/crypto_and_encodings/bcd.dart
t-m-z: Done

lib/widgets/tools/crypto_and_encodings/bcd/bcdlibawcraig.dart
t-m-z: Got the point. But as in base I used a selector list. How do I  reuse only one widget in a selector list?

assets/i18n/de.json
t-m-z: Done - I will check the other selector lists, too.

lib/widgets/tools/crypto_and_encodings/bcd/bcdlibawcraig.dart
S-Man42: As you can see at the segment display section:

You have a base Widget class BCD and the specialized Widget classes. Those can be used in the selector list without any problems. You can have a look at segment selection

lib/widgets/tools/crypto_and_encodings/bcd/bcdlibawcraig.dart
t-m-z: Thanks for this hint

lib/widgets/tools/crypto_and_encodings/bcd/bcdlibawcraig.dart
t-m-z: Done - a steep learning curve for today.

t-m-z:
corrections were done including an update of my local branch

PR #167, opened 2020-08-20
Provide new feature: DTMF

t-m-z:
for review and discussion
tests will follow
probably some indendts have also to be fixed


assets/i18n/de.json
```diff
@@ -822,6 +822,10 @@
   "deletealert_title" : "Wirklich löschen?",
   "deletealert_text" : "Möchtest du wirklich '%s1' löschen?",
 
+  "dtmf_title" : "MFWV",
```
S-Man42, line 825: Is this really the German term? I'd search for DTMF nonetheless: Dualton-Mehrfrequenz-Verfahren... I don't know any Geocacher who doesn't say DTMF... Although Wikipedia agrees your version... I would stay with DTMF. Maybe "DTMF (MFWV)".... Althought... Wikipedia knows DTMF but not MFWV as abbrev. Hmm, I'd go with DTMF and in description: Dualton-Mehrfrequenzwahlverfahren... What do you think?

lib/logic/tools/science_and_technology/dtmf.dart
```diff
@@ -0,0 +1,38 @@
+final DigitToDTMF = {'0' : '1336 941', '1' : '1209 697', '2' : '1336 697', '3' : '1477 697', '4' : '1209 770',
```
S-Man42, line 1: To make it more flexible, I'd take: {'0': [1336,941], ...} (using lists instead of delimited strings). This enables you to simply use the mapSwitch function for the backwards operation instead of listing both permutations.

lib/logic/tools/science_and_technology/dtmf.dart
```diff
@@ -0,0 +1,38 @@
...
+                     '697 1633' : 'A', '770 1633' : 'B', '852 1633' : 'C', '941 1633' : 'D',
+                     '1209 941' : '*', '1477 941' : '#',
+                     '941 1209' : '*', '941 1477' : '#'};
+
+String encodeDTMF(String plain) {
+  return plain
+      .toUpperCase()
+      .split('')
+      .where((digit) => DigitToDTMF[digit] != null)
+      .map((digit) => DigitToDTMF[digit])
```
S-Man42, line 19: Using lists, this line could be changed to 
.map((digit) => DigitToDTMF[digit].join(' ')) // or which delimiting character fits best...

lib/logic/tools/science_and_technology/dtmf.dart
```diff
@@ -0,0 +1,38 @@
...
+      .join('   ');
+}
+
+
+String decodeDTMF(String chiffre) {
+  List chiffreList = chiffre.split(' ');
+  List dtmfCode = new List<String>();
+
+  for(var i = 0; i < chiffreList.length - 1; i = i + 2) {
+    dtmfCode.add(chiffreList[i] + ' ' + chiffreList[i + 1]);
```
S-Man42, line 29: This would be
var tones = [int.tryParse(chiffreList[i]), int.tryParse(chiffreList[i + 1])];
tones.sort();
dtmfCode.add(tones);

Because of the the sort step, it's not necessary to keep both permutations in the list above.

lib/widgets/registry.dart
```diff
@@ -354,6 +355,12 @@ class Registry {
         category: ToolCategory.CRYPTOGRAPHY,
         searchStrings: SEARCHSTRING_ESOTERICPROGRAMMINGLANGUAGE + 'deadfish idso xkcd '
       ),
+      GCWToolWidget(
+        tool: DTMF(),
+        i18nPrefix: 'dtmf',
+        category: ToolCategory.SCIENCE_AND_TECHNOLOGY,
+        searchStrings: 'dual tone multi frequency Mehrfrequenzwahlverfahren Tonwahl'
```
S-Man42, line 362: Currently I stood with lowercase forms. Did you try the capital form. Maybe the search function needs to be empowered, if it isn't already done...

lib/widgets/tools/science_and_technology/dtmf.dart
```diff
@@ -0,0 +1,90 @@
...
+      children: <Widget>[
+        GCWTwoOptionsSwitch(
+          onChanged: (value) {
+            setState(() {
+              _currentMode = value;
+            });
+          },
+        ),
+        _currentMode == GCWSwitchPosition.left
+            ? GCWTextField(
```
S-Man42, line 51: some minor indent issues.

lib/widgets/tools/science_and_technology/dtmf.dart
```diff
@@ -0,0 +1,90 @@
...
+        _buildOutput(context)
+      ],
+    );
+  }
+
+
+  Widget _buildOutput(BuildContext context) {
+    var output = '';
+
+    var textStyle = gcwTextStyle();
```
S-Man42, line 79: That's not quite clear to me... Why did you do this? The normal GCWDefaultOutput already renders the TextDivider, you entered manually, and your used text style is also the default one... Why not simply use GCWDefaultOutput? Am I missing something?

assets/i18n/en.json
```diff
@@ -822,6 +822,10 @@
   "deletealert_title" : "Really remove?",
   "deletealert_text" : "Do you really want to remove '%s1'?",
 
+  "dtmf_title" : "DTMF",
+  "dtmf_description" : "Dual-Tone Multi-Frequenzy",
```
S-Man42, line 826: z -> c :)

lib/logic/tools/science_and_technology/dtmf.dart
```diff
@@ -0,0 +1,38 @@
...
+                     '1209 941' : '*', '1477 941' : '#',
+                     '941 1209' : '*', '941 1477' : '#'};
+
+String encodeDTMF(String plain) {
+  return plain
+      .toUpperCase()
+      .split('')
+      .where((digit) => DigitToDTMF[digit] != null)
+      .map((digit) => DigitToDTMF[digit])
+      .join('   ');
```
S-Man42, line 20: Hm, I see the reason you used this. Although I am not sure, if I had done this, I believe it's a good idea. But maybe you could use a longer space character instead of three spaces. This was already done in Morse, as well: 
https://github.com/S-Man42/GCWizard/blob/1.1.0/lib/logic/tools/crypto_and_encodings/morse.dart#L37

S-Man42: Although I believe the large space character fits well... If you don't like it, I'd prefer the pipe solution because of in-app consistency

lib/logic/tools/science_and_technology/dtmf.dart
S-Man42: Well, yes, I'd go with my solution, it is much more flexible and it spares some code. What exactly was the problem with the refactoring? (We should discuss this in a phone call or something)

lib/logic/tools/science_and_technology/dtmf.dart
t-m-z: Yes, but where is the difference? Of course I have to change some more statements to build the String-output.

lib/logic/tools/science_and_technology/dtmf.dart
t-m-z: I receive an error message:
════════ Exception caught by widgets library ════════════════════════════════════════════
The following NoSuchMethodError was thrown building DTMF(dirty, dependencies: [_LocalizationsScope-[GlobalKey#d4f46]], state: DTMFState#56286):
The method '>' was called on null.
Receiver: null
Tried calling: >(770)

t-m-z: the DTMF-logic-dart-file is part of the PR

t-m-z:
Start working on refactoring

PR #171, opened 2020-08-27
DTMF - need help

t-m-z:
As recommended  I rewrote the logic using a Map with a character/list of int as Key/Value.
Regarding Decoding, I cannato get the Value with a list as key. I always get null.
See line 68


t-m-z:
Problem could be solved. I will open a new PR with the tests and the refactored logic

PR #180, opened 2020-09-11
feature chronogram

Mike-3:


assets/i18n/de.json
```diff
@@ -318,6 +318,11 @@
   "chickenlanguage_description" : "Spielsprache, Geheimsprache",
   "chickenlanguage_example" : "Huhn → huhudefuhn",
 
+  "chronogram_title" : "Chronogramm",
+  "chronogram_description" : "Übersetzt Texte in passenden Wert",
```
S-Man42, line 322: Without the example this wouldn't be clear to me... So, a better explanation should be found here. Maybe it could make sense to create a Roman Numerals Selection for the current Roman function and the Chronogram in it?

lib/logic/tools/crypto_and_encodings/chronogram.dart
```diff
@@ -0,0 +1,27 @@
...
+  String _sortedChars = '';
+
+  input = input.toUpperCase();
+  if (withJU)  {
+    input = input.replaceAll(RegExp(r'[J]'), 'I');
+    input = input.replaceAll(RegExp(r'[U]'), 'V');
+  }
+
+  // disregarded the Roman rule of subtraction and simply added the numerical values
+  _romanToNumber.keys.forEach((romanChar) {
```
S-Man42, line 19: For avoiding code replication, please use the already existing roman function :)

lib/widgets/registry.dart
```diff
@@ -320,6 +321,12 @@ class Registry {
         category: ToolCategory.CRYPTOGRAPHY,
         searchStrings: SEARCHSTRING_GAMELANGUAGE + 'chickenlanguage huehnersprache huenersprache huhn'
       ),
+      GCWToolWidget(
+      tool: Chronogram(),
```
S-Man42, line 325: Funny, negative wrong indents :D That's a new one :D

lib/widgets/tools/crypto_and_encodings/chronogram.dart
```diff
@@ -0,0 +1,52 @@
...
+        ),
+        GCWDefaultOutput(
+            child: _buildOutput()
+        )
+      ],
+    );
+  }
+
+  _buildOutput() {
+    if (_currentInput == null)
```
S-Man42, line 45: This is catched in the logic, so this doesn't need to be done the in frontend. So, if I don't miss anything, you can simply write
```
GCWDefaultOutput(
   child: decodeChronogram(...)
)
```

lib/widgets/tools/crypto_and_encodings/chronogram.dart
```diff
@@ -0,0 +1,52 @@
...
+          title: i18n(context, 'chronogram_modus'),
+          value: _currentValue,
+          onChanged: (value) {
+            setState(() {
+              _currentValue = value;
+            });
+          },
+        ),
+        GCWDefaultOutput(
+            child: _buildOutput()
```
S-Man42, line 38: Indents ;) :D Sorry, it's not that relevant. If it would be the only issue, this wouldn't be something to reject the PR. It's something I always correct if I see it in any file (an there are still lots of indent issues in my own work :D)

S-Man42 (commented):
Nice! :)

Only issue is: Reusing already existing and tested functions to avoid code replication.

assets/i18n/de.json
Mike-3: description changed to "Addition of Roman numerals"
One could make an extended display window. But with the Roman numerals widget, the letters used are not displayed either.

lib/logic/tools/crypto_and_encodings/chronogram.dart
Mike-3: That does not work. This doesn't always produce the same result. Because the order is taken into account with the Roman numerals widget.

lib/widgets/registry.dart
Mike-3: changed

lib/widgets/tools/crypto_and_encodings/chronogram.dart
Mike-3: changed.
copied from ccitt1 ;-)

lib/widgets/tools/crypto_and_encodings/chronogram.dart
Mike-3: changed

lib/widgets/tools/crypto_and_encodings/chronogram.dart
S-Man42: I know... There a many of these... When I find one, I correct instantly... But they're created also with copy/paste by the IDE... Nasty :D

S-Man42:
The Destination Branch is still 1.1.0. I guess, you need to close the PR and create a new one.

PR #183, opened 2020-09-14
Bacon chiffre extension (inverse, binary)

Mike-3:


S-Man42:
Fixed Typo and used internal substitution method for avoiding code redundancy :)

PR #191, opened 2020-09-22
Kenny Code (Ignore Case Sensitivity)

Mike-3:


test/logic/tools/crypto_and_encodings/kenny.dart
```diff
@@ -31,43 +31,80 @@ void main() {
...
+      {'input' : 'pppmppmffppp pppppfmmfmfp pppmffmmfmfpfmp', 'replaceCharacters': ['m', 'p', 'f'], 'expectedOutput' : 'NEIN NOCH NICHT', 'caseSensitive' : true},
+      {'input' : 'pspspsmfgpspsmfgfd1fd1pspsps pspspspspsfd1mfgmfgfd1mfgfd1ps pspspsmfgfd1fd1mfgmfgfd1mfgfd1psfd1mfgps', 'replaceCharacters': ['mfg', 'ps', 'fd1'], 'expectedOutput' : 'NEIN NOCH NICHT', 'caseSensitive' : true},
+      {'input' : 'pmpmppppppppffm mmfppfmpmmpp mmmfmfmmfmfp fmmmmfmfpppfppp mpmpffmffppp?', 'replaceCharacters': ['m', 'p', 'f'], 'expectedOutput' : 'KENNY CODE AUCH SCHON DRIN?', 'caseSensitive' : true},
+      {'input' : 'mmmas', 'replaceCharacters': ['m', 'p', 'f'], 'expectedOutput' : 'AAS', 'caseSensitive' : true},
+      {'input' : 'mmmmas', 'replaceCharacters': ['m', 'p', 'f'], 'expectedOutput' : 'AMAS', 'caseSensitive' : true},
+
+      {'input' : 'mpfapff', 'replaceCharacters': ['m', 'p', 'f'], 'expectedOutput' : 'FAR', 'caseSensitive' : true},
+      {'input' : 'Mfmfmffmp mfmmppppmmmmmmfmfpfmp!', 'replaceCharacters': ['m', 'p', 'f'], 'expectedOutput' : 'MUIMP GEMACHT!', 'caseSensitive' : true},
```
S-Man42, line 66: When I see correctly, this is the only test case which includes upper case characters, right? I'd include some more simple ones in case you need them for debugging:

M
MF
MFM
Mf
Mfm
Mfmfmf
MfmFmf
....

S-Man42 (commented):
Good code. Is there a reason why you didn't made it for encryption, too?

Mike-3:
> Good code. Is there a reason why you didn't made it for encryption, too?

I also thought about it, but what should be in the output? In the configuration field there is, for example, 'm', should then 'M' appear in the output? Doesn't really make sense, I think.
Then it is not an Ignore Case-Sensitive, rather a toUpperCase.

S-Man42:
I think, if the character is lower case, the 3 encryption characters are lower case, otherwise the first of the three character is capital. As it is at the example code above... I just saw this variant sometimes. I believe it's a reasonable way

PR #192, opened 2020-09-22
RC4

Mike-3:


assets/i18n/en.json
```diff
@@ -1301,6 +1301,14 @@
   "railfence_optionalpassword" : "Optional Password",
   "railfence_offset" : "Offset",
 
+  "rc4_title" : "RC4",
+  "rc4_description" : "stream cipher (HTTPS, SSH 1, WEP, WPA)",
+  "rc4_example" : "abcdef → CD14",
+  "rc4_format" : "Format",
+  "rc4_format_auto" : "Auto",
+  "rc4_format_text" : "Text",
+  "rc4_format_asciivalues" : "ASCII values",
```
S-Man42, line 1310: I'd prefer "ASCII Values" with a capital V for consistency purposes :)

lib/logic/tools/crypto_and_encodings/rc4.dart
```diff
@@ -0,0 +1,181 @@
...
+class RC4Output {
+  final String output;
+  final ErrorCode errorCode;
+
+  RC4Output(this.output, this.errorCode);
+}
+
+RC4Output useRC4(String input, InputFormat inputFormat, String key, InputFormat keyFormat, OutputFormat outputFormat){
+  if (input == null || input == '')
+    return new RC4Output('', ErrorCode.OK);
```
S-Man42, line 17: No huge issue: But in the entire app I decided to leave the "new" keyword. It's not wrong or a problem but in case you'd touch this file again, you could remove this for consistency purposes?

lib/logic/tools/crypto_and_encodings/rc4.dart
```diff
@@ -0,0 +1,181 @@
...
+  switch (format) {
+    case InputFormat.AUTO:
+      out = convertInputToIntList(input, InputFormat.BINARY);
+      if (out.length == 0)
+        out = convertInputToIntList(input, InputFormat.ASCIIVALUES);
+      if (out.length == 0) {
+        out = convertInputToIntList(input, InputFormat.HEX);
+        if (out.length > 0) {
+          for (var item in out) {
+            // to large ?
```
S-Man42, line 118: typo. Should be "too", I guess

lib/widgets/registry.dart
```diff
@@ -526,6 +527,12 @@ class Registry {
         category: ToolCategory.CRYPTOGRAPHY,
         searchStrings: 'railfence lattenzaun jaegerzaun zigzag redefence zig-zag palisadenzaun gartenzaun transposition'
       ),
+      GCWToolWidget(
+          tool: RC4(),
```
S-Man42, line 531: Indent issue :)

lib/widgets/registry.dart
```diff
@@ -526,6 +527,12 @@ class Registry {
...
+          i18nPrefix: 'rc4',
+          category: ToolCategory.CRYPTOGRAPHY,
+          searchStrings: 'rc4 arc4 arcfour stream cipher stromverschluesselung https ssh ssl wep wpa'
```
S-Man42, line 534: It's better to write "streamcipher" instead of "stream cipher". Currently, if someone searches for "streamcipher", he wouldn't be successful. On the other hand "streamcipher" also supports the search string "stream cipher" automatically.

lib/widgets/tools/crypto_and_encodings/rc4.dart
```diff
@@ -0,0 +1,168 @@
...
+  Widget _buildOutput(BuildContext context) {
+    if (_currentInput == null || _currentInput.length == 0) {
+      return GCWDefaultOutput();
+    }
+
+    var _currentOutput = useRC4(_currentInput, _currentInputFormat, _currentKey, _currentKeyFormat, _currentOutputFormat);
+
+    if (_currentOutput == null || _currentOutput.errorCode != ErrorCode.OK) {
+      return GCWDefaultOutput(
+        // child: _currentOutput.errorCode.toString(),
```
S-Man42, line 159: Please remove unneccessary comments :)

lib/logic/tools/crypto_and_encodings/rc4.dart
```diff
@@ -0,0 +1,181 @@
...
+      out = _convertToIntList(input, 10);
+      break;
+    default:
+      break;
+  }
+
+  return out;
+}
+
+List<int> _convertToIntList(String input, int base) {
```
S-Man42, line 148: I am not quite sure if this does the same... However, please check if you could use the textToIntList() or similar functions from the global common_utils instead. if not, could this moved to be used as a more global function?

lib/logic/tools/crypto_and_encodings/rc4.dart
```diff
@@ -0,0 +1,181 @@
...
+    s[i] = s[j];
+    s[j] = temp;
+    randomValue = s[(s[i] + s[j]) % 256];
+    out.add(randomValue ^ input[n]);
+  }
+
+  return out;
+}
+
+List<int> convertInputToIntList(String input, InputFormat format) {
```
S-Man42, line 103: I guess, this could be a private function?

lib/logic/tools/crypto_and_encodings/rc4.dart
```diff
@@ -0,0 +1,181 @@
...
+enum ErrorCode{OK, INPUT_FORMAT, KEY_FORMAT, MISSING_KEY}
+
+class RC4Output {
+  final String output;
+  final ErrorCode errorCode;
+
+  RC4Output(this.output, this.errorCode);
+}
+
+RC4Output useRC4(String input, InputFormat inputFormat, String key, InputFormat keyFormat, OutputFormat outputFormat){
```
S-Man42, line 15: The method name is quite uncommon in the app. I'd prefer "encryptRC4" instead, what do you think?

lib/logic/tools/crypto_and_encodings/rc4.dart
```diff
@@ -0,0 +1,181 @@
...
+        break;
+      default:
+        break;
+    }
+  });
+
+  return new RC4Output(out, ErrorCode.OK);
+}
+
+List<int> rc4(List<int> input, List<int> key){
```
S-Man42, line 69: Does this method need to be public?

S-Man42 (commented):
In general good code, as usual! Well done! 

Nevertheless, there are some minor issues and some issues to be discussed. But they shouldn't make big troubles :)

S-Man42:
Uuh, I see something problematic. I took the GCC with the same inputs, but GCWizard's output is not the same unfortunately :(

Input: ABC, Key: 1234, both Auto
Result GCWizard: (Hex): A63
Result GCC: (Hex): 09 9d b7

Which one is incorrect?

lib/widgets/tools/crypto_and_encodings/rc4.dart
Mike-3: Are short popup messages possible? As with the GCC. for example key is missing. That is what it was intended for.

lib/logic/tools/crypto_and_encodings/rc4.dart
Mike-3: Decrypting and encrypting are the same function, applied the same twice, gives the origin. Hence the name. I think cryptRC4 is good (from GCC).

Mike-3:
> 
> 
> Uuh, I see something problematic. I took the GCC with the same inputs, but GCWizard's output is not the same unfortunately :(
> 
> Input: ABC, Key: 1234, both Auto
> Result GCWizard: (Hex): A63
> Result GCC: (Hex): 09 9d b7
> 
> Which one is incorrect?

Good question. GCC means "ABC" is a text and GC Wizard means it is a hex number. That is probably the "problem" of the automatic.
GC Wizard has the same result when I select Text when entering.

lib/widgets/tools/crypto_and_encodings/rc4.dart
S-Man42: Yes, it's possible using the "showToast()" method. However, error handling is currently a missing feature in the entire app. @t-m-z and I, we discussed this earlier, but found no good way to implement this in a global way as it should be done. Maybe you both could discuss this any further.

But in nearly all tools throughout the app, possible errors are catched internally, either by ignore bad inputs or output empty strings or similar. So, error messages could be avoided in general - some exceptions exists. 

However, if you really need an error message, just use the toast. But I'd prefer using complete, readable error messages instead of error codes. The users do not know how to interpret codes and we have no global dictionary for a consistent coding.  Have a look at RSA to see how they can be handled - throwing Exception with message, catching and showing them in a toast message.

lib/logic/tools/crypto_and_encodings/rc4.dart
S-Man42: ok.

Mike-3:
I discovered the difference. GCC only accepts hex numbers with 2 characters. I can change the code.

S-Man42:
That sounds more technical (as binary representation with fixed 8 characters...).I am not quite sure about it... Maybe have a look at other online tools... How do they behave?

test/logic/tools/crypto_and_encodings/rc4.dart
```diff
@@ -12,16 +12,17 @@ void main() {
...
+      {'input' : 'ABC', 'format' : InputFormat.AUTO, 'key' : '1234', "keyFormat" : InputFormat.AUTO, 'expectedOutput' : '09 9D B7', 'outputFormat' : OutputFormat.HEX, 'errorCode' : ErrorCode.OK},
 
-      {'input' : 'ABCDEF', 'format' : InputFormat.AUTO, 'key' : 'FEDCBA', "keyFormat" : InputFormat.AUTO, 'expectedOutput' : 'E3 2D 67 E0 CF FA', 'outputFormat' : OutputFormat.HEX, 'errorCode' : ErrorCode.OK},
-      {'input' : 'ABCDEF', 'format' : InputFormat.AUTO, 'key' : 'FE', "keyFormat" : InputFormat.AUTO, 'expectedOutput' : '72 55 5C DD 7D 1E', 'outputFormat' : OutputFormat.HEX, 'errorCode' : ErrorCode.OK},
-      {'input' : 'ABCDEF', 'format' : InputFormat.AUTO, 'key' : 'FE', "keyFormat" : InputFormat.AUTO, 'expectedOutput' : 'rU\\Ý}\x1E', 'outputFormat' : OutputFormat.TEXT, 'errorCode' : ErrorCode.OK},
+      {'input' : 'ABCDEF', 'format' : InputFormat.AUTO, 'key' : 'FEDCBA', "keyFormat" : InputFormat.AUTO, 'expectedOutput' : 'ABCDD1', 'outputFormat' : OutputFormat.HEX, 'errorCode' : ErrorCode.OK},
```
S-Man42, line 23: Why doesn't the HEX output contain spaces while it does in line 21?

lib/logic/tools/crypto_and_encodings/rc4.dart
```diff
@@ -0,0 +1,193 @@
...
+      });
+
+    if (ok)
+      return InputFormat.ASCIIVALUES;
+  }
+
+  return InputFormat.TEXT;
+}
+
+List<int> convertToIntList(String input, int base) {
```
S-Man42, line 160: Please make this private again

test/logic/tools/crypto_and_encodings/rc4.dart
Mike-3: Input ABC is recognized as text and broken down into 3 individual letters. Therefore 3 output values. The recognition routine expects an even number of HEX characters. ABCDEF is recognized as a hex number because of the even number of characters.
Has been adopted by the GCC.
But that doesn't really make sense.
Perhaps you should only search for HEX characters or limit it to groups of two. If more than 2 characters, it is not a HEX value. Unless you explicitly choose HEX.

PR #206, opened 2020-10-01
feature: Gade

t-m-z:
tests were passed successfully


lib/logic/tools/crypto_and_encodings/gade.dart
```diff
@@ -0,0 +1,24 @@
+//http://www.crumlin.dk/geocaching/gade/#:~:text=A%20Gade%20is%20a%20method%2C%20invented%20by%20the,least%20one%20variable%20containing%20each%20digit%20%28including%20zero%29.
+
+import 'package:flutter/material.dart';
```
S-Man42, line 3: Please avoid using flutter related imports in the logic

lib/logic/tools/crypto_and_encodings/gade.dart
```diff
@@ -0,0 +1,24 @@
...
+
+String buildGade(String input){
```
S-Man42, line 5: As output I'd prefer a map:
{'A': 1, 'B': 2, 'C': 3, ...}

It's much easier to handle the output later.

lib/widgets/main_view.dart
```diff
@@ -88,6 +88,7 @@ import 'package:gc_wizard/widgets/utils/common_widget_utils.dart';
 import 'package:prefs/prefs.dart';
 import 'package:url_launcher/url_launcher.dart';
 
+
```
S-Man42, line 91: Please remove this line. It's not necessary to change the main_view in this feature.

assets/i18n/en.json
```diff
@@ -317,6 +317,15 @@
   "changelog_0.0.4" : "[fix] Calculation of hours/minutes/seconds in 'Count Days'",
   "changelog_0.0.1" : "[new] Google PlayStore Initialisation",
 
+  "chao_title" : "Chao Cipher",
```
S-Man42, line 320: Will the CHAO string be added in CHAO PR as well? Normally, they need to be removed here, as they are not related to Gade

lib/widgets/tools/crypto_and_encodings/gade.dart
```diff
@@ -0,0 +1,50 @@
...
+      children: <Widget>[
+        GCWTextField(
+          controller: _GadeInputController,
+          onChanged: (text) {
+            setState(() {
+              _currentGadeInput = text;
+            });
+          },
+        ),
+        GCWTextDivider(
```
S-Man42, line 41: 1. This code could use the GCWDefaultOutput instead, which creates the TextDivider and OutputText by itself :)

2. Nonetheless I'd prefer to create the output as columnedMultilineOutput (as seen in many other tools, like astronomy, periodic tables, permutation, colors, ...): 
```

     GCWDefaultOutput(
          child: Column(
            children: columnedMultiLineOutput(
              yourOutputMap.entries.map((entry) {
                return [entry.key, entry.value];
              }).toList()
            ),
          ),
        )
```

assets/i18n/en.json
t-m-z: done - I can not remember to have this added. Probably a Pull went wrong.

lib/logic/tools/crypto_and_encodings/gade.dart
t-m-z: done - also I am sure not to have added this line.

lib/widgets/main_view.dart
t-m-z: done

S-Man42:
Because of the many other accepted PRs, some conflicts in the i18n files arose. Please check this before committing :)

t-m-z:
to resolve the conflicts and change the output I will start a new pull request from the scratch

lib/logic/tools/crypto_and_encodings/gade.dart
t-m-z: done

lib/widgets/tools/crypto_and_encodings/gade.dart
t-m-z: done

PR #213, opened 2020-10-02
feature: Chao Cipher

t-m-z:
The Chaocipher is a cipher method invented by John Francis Byrne in 1918
see also https://www.dcode.fr/chao-cipher


S-Man42:
Merged but with following issue #221

PR #218, opened 2020-10-08
AMSCO Cipher

Mike-3:


S-Man42:
1. I checked a few online tools. All of them start with 1 letter. So the grid will be filled with 1-2-1-2-... You fill the grid other way round: 2-1-2-1-2-..., which, naturally, creates another output. Is this a bug or intended?

2. All reference implementations I checked remove spaces and special chars, whereas you are using it. Furthermore all of them convert the cipher output to upper case. I'd prefer to adapt this behavior, what to you think?

3. I'd prefer to suppress the "Used Grid" divider on empty output, as it is done in Polybios.

Mike-3:
1/ 2.
http://kryptografie.de/kryptografie/chiffre/amsco.htm (my reference)
- no UC, no remove spaces (only recommendation)
- start with 2 characters
https://www.dcode.fr/amsco-cipher
- UC, remove spaces
- start with 1 characters
https://encode-decode.appspot.com/cipher_forms/amsco.html
- UC, , no remove spaces
- Option. Start with one or two characters
there are just different implementations. I was of the opinion that you are most flexible with the current one, because you can do that with capital letters and spaces before encryption.

Suggestion: I would leave UC and spaces as they are for flexibility. But start with a sign. Can also be optionally changed later. Or we make an option (is my favorite).

3. OK

4. Ich bin für AMSCO. Ist glaube ich auch für eine Abkürzung. Wird meistens auch so geschrieben. Ist auch so in den Sprachfiles.

Mike-3:
Correction:
https://encode-decode.appspot.com/cipher_forms/amsco.html
- remove spaces (Output in groups of 5 letters, the error was due to my short test text)

More implementations with no spaces and capital letters.
I am changing it.

Issue #177, opened 2020-09-06
Numeral words


S-Man42:
(from Geo-Link)

German: Zahlwörter

A simple tool with a dropdown button for choosing a language. After choosing it shows numeral words in this language:

German:
1 Eins
2 Zwei
3 Drei

English
1 One
2 Two
3 ...
------------------------------------
Enhanced version of the tool: Simple Text input, which searches for word. So, when typing "eins", it shows 1 (German), Maybe it could be possible to type "eins two quatre" which results in:

Output:
124
Detailed Output:
1: German
2: English
3: French
----------------------------
The dictionary data should be stored as maps:

{
'i18n_key_for_german': {1: 'eins', 2: 'zwei', ...},
'18n_key_for_english' : {1: 'one', 10: 'ten'}
}

(key should be 'common_language_german', 'common_language_english', ...)

Each language should contain at least the numbers 0 to 10.

PR #232, opened 2020-10-13
feature: Numeral words

t-m-z:
The feature provides two basic functions:

- lists numeral words for given languages: DEU, ENG, FRA, ITA, ESP, POR, DNK, SWE, NOR, NLD, RUS, RUS (Cyrillic), POL, Volapük

- analyzes a given text to identify numeral words with a specific language or with all languages

1.  analyzes only whole words e.g. eighthundred and one => 801

1. analyzes also part of words e.g. weight => eight => 8

be careful - there could be still some errors in the logic


S-Man42:
As promised: Added the commit for restructured widgets and many test cases that don't run properly.

Some thoughts:
- You should't really make a difference between "entire" and "parts" in you logic. If "parts" run properly, it should include the entire word: "ABC" is part of "ABC". This is currently not the case, it seems that you are using different logic implementations for both cases. Or maybe I missed something ?!
- English is quite hard: it is nearly not decideable what output one expect when entering "hundred fifty". Is it "150" or "100 50". I guess, you have to decide this and implement it consistently. So, maybe you have to change the test cases to your interpretation

- Output: I would like to see 2 changes: 
1. Make two outputs: One only the decoded numbers. It's better to copy the result and put it into the coords function or something
2. Your current output as detailed description/explanation. But I would like to see this as 3-column multiLinedOutput: Found number, the related numeral word (as it is currently) and for chosen "all languages" addionally the language. I had a look at your code. Unfortunately you didn't keep all these information separately and put it into a string on-the-fly. I'd like to change this. So, if you don't want to change this by yourself, I offer you to refactor this after you fixed the logic. This is more important: Form follows function.

t-m-z:
regarding entire/parts there is a different logic because I do not parse for every possible numeral word. I want to identify in general only the digits 0-9.
Regarding English/entire words I believe it is necessary to stick to the official grammar.

Regarding the test cases, I need some help: I have no glue how to provide the expected output.

S-Man42:
#177 done.

PR #234, opened 2020-10-22
Substitution Breaker

Mike-3:


S-Man42:
I close this and make a new one from my own branch which includes the export button the substitution tool. Unfortunately I couldn't commit it directly to this branch because of missing write access

PR #262, opened 2020-11-21
feature: Burrows Wheeler Transformation

t-m-z:
http://kryptografie.de/kryptografie/chiffre/burrows-wheeler.htm
https://www.geocachingtoolbox.com/index.php?lang=en&page=burrowsWheelerTransform

https://de.wikipedia.org/wiki/Burrows-Wheeler-Transformation


lib/widgets/tools/crypto_and_encodings/burrows_wheeler.dart
```diff
@@ -0,0 +1,118 @@
...
+          onChanged: (value) {
+            setState(() {
+              _currentMode = value;
+            });
+          },
+        ),
+        _currentMode == GCWSwitchPosition.left
+        ? GCWTextField(
+            controller: _plainController,
+            hintText: i18n(context, 'burrowswheeler_inputplain'),
```
S-Man42, line 65: Hint text is not shown in my emulator

lib/widgets/registry.dart
```diff
@@ -346,6 +347,12 @@ class Registry {
         category: ToolCategory.CRYPTOGRAPHY,
         searchStrings: SEARCHSTRING_BRAINFK
       ),
+      GCWTool(
+        tool: BurrowsWheeler(),
+        i18nPrefix: 'burrowswheeler',
+        category: ToolCategory.CRYPTOGRAPHY,
+        searchStrings: 'burrows wheeler'
```
S-Man42, line 354: better: burrows-wheeler-transformation burrowswheelertransformation bwt bzip2

assets/i18n/de.json
```diff
@@ -297,6 +297,12 @@
   "brainfk_error_loopnotopened" : "Schleife nicht geöffnet",
   "brainfk_ook_title" : "Ook",
 
+  "burrowswheeler_title" : "Burrows-Wheeler Transformation",
```
S-Man42, line 300: maybe add something with bzip2 into the description? :)

S-Man42 (commented):
It is not quite clear, how to use this tool. For Index I can only enter #~| characters. But using the Wikipedia (EN/DE) example: ".NNAAA.S" this results in nothing. Moreover, the Wikipedia (and the error message) states that you need an index, which seems to be an integer. But I cannot enter an integer value. When I read the Wikipedia article, I would assume, that there's a text field for the plain/crypto text and an integer spinner for the index. Maybe I didn't understand the tool?

t-m-z:
Regarding your comments - thanks first for your hints.

The search-strings are added, also the de.json and en.json are corrected that the hints are now shown.

Regarding the usage: I focused on the description from kryptographie.de. Hence, there are two ways to concider:
after the encoding you have the encodes text and an index. This index is an integer which could be stored within the text. Therefore you need a special character, which does not belong to the characters of the text.
For decoding you need either the index as a numeric value or the special character which represents the index.

The tests will be adapted as the next step.

S-Man42:
http://fbcs.bplaced.net/multi_encoder_decoder.html -> DVORAK I, DVORAK II US, DVORAK II DE

t-m-z:
all tests were successful
the frontend provides the encryption or decryption of a given text with 

- option: encrypt the index; hence you have to provide a symbol, which is not a symbol from the plaintext. This symbol will be placed at the index-position. Otherwise you will get an index-number

- option: compress the output; hence characters from the same kind will bei replced by a digit. Therefore all digits within the plaintext will be replaced by numeral words in Eenglish or German language

assets/i18n/en.json
```diff
@@ -297,6 +297,22 @@
...
+  "burrowswheeler_language" : "Language",
+  "burrowswheeler_language_deu" : "German",
+  "burrowswheeler_language_eng" : "English",
+  "burrowswheeler_index" : "Index",
+  "burrowswheeler_index_encode" : "Encrypt index",
+  "burrowswheeler_index_symbol" : "Index-Symbol",
+  "burrowswheeler_inputplain" : "Plaintext",
+  "burrowswheeler_inputchiffre" : "Chiffretext",
+  "burrowswheeler_error_no_index" : "ERROR: Index is missing",
+  "burrowswheeler_error_char_index" : "ERROR: Index-Character should not be part of text",
```
S-Man42, line 313: "Index character" :)

assets/i18n/en.json
```diff
@@ -297,6 +297,22 @@
...
+  "burrowswheeler_compress_output" : "Compressed output",
+  "burrowswheeler_decompress_output" : "Decompressed output",
+  "burrowswheeler_language" : "Language",
+  "burrowswheeler_language_deu" : "German",
+  "burrowswheeler_language_eng" : "English",
+  "burrowswheeler_index" : "Index",
+  "burrowswheeler_index_encode" : "Encrypt index",
+  "burrowswheeler_index_symbol" : "Index-Symbol",
+  "burrowswheeler_inputplain" : "Plaintext",
+  "burrowswheeler_inputchiffre" : "Chiffretext",
```
S-Man42, line 311: "Plain text" and "Cipher text". Could you please fix that in "Chao" as well? :)

assets/i18n/en.json
```diff
@@ -297,6 +297,22 @@
...
+  "burrowswheeler_title" : "Burrows-Wheeler Transformation",
+  "burrowswheeler_description" : "Basis for bzip-compression",
+  "burrowswheeler_compress_output" : "Compressed output",
+  "burrowswheeler_decompress_output" : "Decompressed output",
+  "burrowswheeler_language" : "Language",
+  "burrowswheeler_language_deu" : "German",
+  "burrowswheeler_language_eng" : "English",
+  "burrowswheeler_index" : "Index",
+  "burrowswheeler_index_encode" : "Encrypt index",
+  "burrowswheeler_index_symbol" : "Index-Symbol",
```
S-Man42, line 309: Index symbol

lib/logic/tools/crypto_and_encodings/burrows_wheeler.dart
```diff
@@ -0,0 +1,170 @@
+import 'package:gc_wizard/utils/common_utils.dart';
+
+final Map DEdigitToString = {0 : 'null', 1 : 'eins', 2 : 'zwei', 3 : 'drei', 4 : 'vier', 5 : 'fuenf', 6 : 'sechs', 7 : 'sieben', 8 : 'acht', 9 : 'neun', };
+final Map ENdigitToString = {0 : 'zero', 1 : 'one', 2 : 'two', 3 : 'three', 4 : 'four', 5 : 'five', 6 : 'six', 7 : 'seven', 8 : 'eight', 9 : 'nine', };
```
S-Man42, line 4: Could you re-use the strings from your numeral words dictionary?

Furthermore: On Encryption you could encrypt DE and EN if you find it, couldn't you? That woudl spare a switch.

lib/logic/tools/crypto_and_encodings/burrows_wheeler.dart
```diff
@@ -0,0 +1,170 @@
...
+      .replaceAll(language[2], '2')
+      .replaceAll(language[3], '3')
+      .replaceAll(language[4], '4')
+      .replaceAll(language[5], '5')
+      .replaceAll(language[6], '6')
+      .replaceAll(language[7], '7')
+      .replaceAll(language[8], '8')
+      .replaceAll(language[9], '9');
+
+return BWTOutput(decodiert, indexChar, decompressed);
```
S-Man42, line 169: missing indent

lib/logic/tools/crypto_and_encodings/burrows_wheeler.dart
```diff
@@ -0,0 +1,170 @@
...
+  String decodiert = '';
+  for (int i = 0; i < len; i++){
+    if (tabelle[index] != null) {
+      decodiert = decodiert + tabelle[index][1];
+      index = int.parse(tabelle[index][0]);
+    }
+  }
+
+  if (compress)
+    decodiert = decodiert
```
S-Man42, line 157: Is this really working correctly? What if input really was "123". The decryption would result in "onetwothree" as I can see, wouldn't it?

t-m-z:
json corrected, compression removed, test adjusted

S-Man42:
Adjusted the Widget a little bit and removed some "compression" artifacts from tests and logic. Please accept my PR into your feature branch: https://github.com/t-m-z/GCWizard/pull/3

I believe I have still some problems with the logic. I tried the Wikipedia example (https://de.wikipedia.org/wiki/Burrows-Wheeler-Transformation):
Expected: ".NNAAA.S", index 7
Actual: "S..NNAAA", index 2

Furthermore, decoding the Wikipedia example was not possible for me:
Input ".NNAAA.S", index 7
Expected: ".ANANAS."
Actual: "NAAAN.NA" (note, that the S and one dot is missing! Instead there's an additional N and an A)

Last but not least you get a RedScreenOfDeath when:
1. Decode
2. Input some text, e.g. ".NNAAA.S"
3. Switch to "Index Symbol"
4. Remove the initial "#" character
5. Enter dot -> CRASH

t-m-z:
Thanks for your help 
- PR accepted
- max index is set correct - equal to length of plain/-ciphertetxt
- regarding the Wikipedia-Example .ANANAS. - this does not work with kryptografie.de, too. However, the example Text: „a!iepdWkii“, Index: 2 works fine. The explanation is easy: it is the alphabetical sorting: Wikipedia sorts the dot after the "A", but ASCII sorts the "A" after the dot. If you replace the dot with ^, the example is correct.
Hence I made a change request for the Wikipedia article.
- I added the wikipedia examples
- regarding the RedSCreenOfDeath - I have no clue what happend. The widget checks, if the symbol is null or empty and provides an error-message.

PR #264, opened 2020-11-22
gpx/kml export

Mike-3:


Mike-3:
https://github.com/cgeo/cgeo/issues/6964
https://github.com/cgeo/cgeo/issues/527
https://stackoverflow.com/questions/64970460/open-kml-gpx-files-in-appropriate-external-map-app

PR #272, opened 2020-12-02
GC wizard symbol tables pdf (symbol_table revision)

Mike-3:
with special symbol list from dart file


S-Man42:
There are some issues at the first scan of the German version (I guess, there are more, but these can be used as indices, I guess):
- aurebesh, futurama, hazard: quotation, quotation_in, quotation_out, underscore missing
- klingon kli: TLH, GH, ... should be small letters
- cyrillic: not used special map from config.json
- suetterlin: MM must be mm, small äöü not replaced

- version number missing in document
- sorting differs to tool: Translated symbols to the end (see semaphore)

Addional notes:
- Please increase the font size. Printed on a A4 document, Grandma Erna will not read the letters anymore, I guess
- In the app, for space I used the space sign ␣ to avoid showing nothing. Could you add this as well?

Issue #263, opened 2020-11-22
Coordinates Intersections of Three Circles Crash


Mike-3:
Crashed with 2 circles with the same center and diameter.

Mike-3:
The same with Intersections of Two Circles

S-Man42:
I cannot reproduce this. Can you explain the exact workflow. What exactly crashed?

Mike-3:
At that moment it worked the first time. The result was no cut points. After that, I've had it twice in a row.
The app is then frozen.
![20201123_151940](https://user-images.githubusercontent.com/70719484/99973946-9a334280-2da0-11eb-90da-dfbcaf6999e2.jpg)
![20201123_152009](https://user-images.githubusercontent.com/70719484/99973962-9c959c80-2da0-11eb-8d45-f6ecd4e7efc1.jpg)

S-Man42:
Ah, same circles... This can be. I have to check this.

S-Man42:
Done #275

PR #275, opened 2020-12-11
Bugfix circle intersections

Mike-3:


Mike-3:
Issues ##265, #263

lib/logic/tools/coords/intersect_two_circles.dart
```diff
@@ -15,6 +15,10 @@ class _IntersectTwoCirclesCalculator extends IntervalCalculator {
     var r1 = parameters['radius1'];
     var r2 = parameters['radius2'];
 
+    // same position and diameter
+    if ((r1 == r2) && (distanceToCoord1.a == distanceToCoord2.a) & (distanceToCoord1.b == distanceToCoord2.b))
```
S-Man42, line 19: This is the correct approach but unfortunately the wrong position in code. This function is called (for correct input) many thousand times. So, the check will be done many times without a reason. It would be better to do this check just once before starting the complex algorithm. This place is only few line below: At the top of the intersectTwoCircles Method:
```
if (circlesAreSame) // your code
return [];
```

PR #276, opened 2020-12-12
feature: additional Units

t-m-z:
includes a new category: Typography


t-m-z:
delete Landleague, Leuge - to many historic variants
change Hathom to Nautischer Faden - to many variants for DEU Klafter
change TL, EL to US tsp, tbsp - specific measure
add US Saltspoon and dessertspoon, cup, teacup

https://de.wikipedia.org/wiki/Nautischer_Faden
https://de.wikipedia.org/wiki/Angloamerikanisches_Ma%C3%9Fsystem#K%C3%BCche

assets/i18n/de.json
```diff
@@ -785,6 +814,12 @@
   "common_unit_volume_uspk_name" : "US Peck",
   "common_unit_volume_usbu_name" : "US Bushel",
   "common_unit_volume_bathtub_name" : "Badewanne (150l)",
+  "common_unit_volume_tsp_name" : "Teelöffel (US)",
+  "common_unit_volume_tbsp_name" : "Esslöffel (US",
+  "common_unit_volume_ssp_name" : "Salzlöffel (US)",
+  "common_unit_volume_dsp_name" : "Dessertlöffel (US",
```
S-Man42, line 820: Typo :) Parantheses missing ;)

Do you like to add the German Versions as well? :)

t-m-z: I do not think so because there is no standard. That's the difference between US and DEU.
Older EL should be 15ml, newer EL should be 10ml and TL is something between 2 and 5 ml.
I remember the descussion  ;-)

S-Man42:
Link Unit value is wrong. You have to add a "1". Please check all other values twice :)

t-m-z:
checked according to 
https://physics.nist.gov/cuu/pdf/sp811.pdf 
https://en.wikipedia.org/wiki/Cooking_weights_and_measures
https://trustconverter.com/

PR #283, opened 2020-12-22
feature: Number sequences

t-m-z:
provides calculations on number sequences:
- get - nth number of a sequence
- check if a number belons to a sequence, returns index or -1
- get numbers in a given range
- get numbers with a given length 
- check if a sequence of digits belongs to number, return number, index within the number sequence and position of digits or -1, if not

Number sequences are
-  https://oeis.org/A000129   Pell
-  https://oeis.org/A001045   Jacobsthal
-  https://oeis.org/A000032   Lucas
-  https://oeis.org/A000045   Fibonacci
-  https://oeis.org/A000225   Mersenne
-  https://oeis.org/A000051   Fermat
-  https://oeis.org/A014551   Jacobsthal-Lucas
-  https://oeis.org/A002203   Pell-Lucas
-  https://oeis.org/A005132   Recaman
-  https://oeis.org/A000108   Catalan


S-Man42:
I am very curious about this one. Unfortunately I cannot build this branch. There are many old artifacts (old symbols, old GCWToolWidget name, ...) Please Merge current 1.3.0

lib/widgets/tools/science_and_technology/number_sequences/numbersequences_check.dart
```diff
@@ -0,0 +1,60 @@
...
+        ),
+        GCWIntegerSpinner(
+          value: currentInputN,
+          onChanged: (value) {
+            setState(() {
+              currentInputN = value;
+            });
+          },
+        ),
+        GCWTextDivider(
```
S-Man42, line 46: Please don't use extra Divider if not necessary. The GCWOutput Widgets usually offer all this out-of-the-box. Here you can use 

GCWDefaultOutput(
    child: checkNumber(...)
)

lib/widgets/tools/science_and_technology/number_sequences/numbersequences_contain.dart
```diff
@@ -0,0 +1,76 @@
...
+          min: 0,
+          max: 999999999,
+          onChanged: (value) {
+            setState(() {
+              currentInputN = value;
+            });
+          },
+        ),
+        GCWTextDivider(
+            text: i18n(context, 'common_ouput')
```
S-Man42, line 50: Concerning the previous comment ;) Here you have a typo ("ou_put") which yields a "null" output in the divider in most of your tools. This can be avoided when using the default GCWOutputs ;)

lib/logic/tools/science_and_technology/number_sequence.dart
```diff
@@ -0,0 +1,901 @@
...
+          found = true;
+        } else
+          index++;
+      }
+      break;
+    case NumberSequencesMode.MERSENNE:
+      number = Zero;
+      index = 0;
+      found = false;
+      while (number < checkNumber + One && !found) {
```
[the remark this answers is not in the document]
t-m-z: I understand that point - but this will result in performance issues.
Calculating with BigInt means, that many formulas will not work because of the sqrt-Function.
Also Recaman has to be build with recursion

t-m-z:
thanks for your hints regarding the UI - I believe a kind of documentation about the widgets could be helpful   ;-)
- merge of 1.3.0 done
- refactor widgets done
- refactor json done
- use external formulas done besides Recaman, Lucas, Pell, Fibonacci - but this is not suitable for all functions I believe

PR #284, opened 2020-12-22
Feature Esoteric beatnik

t-m-z:
for discussion, I believe the Frontend is a little bit crowdy and also the code has to be beautified - remove prints etc.
I will do that as soon as possible after implementing your hints.

Also having seen Mike's Input Dialog for whitespace this is something useful to add


Mike-3:
Is really a lot of expense.
With Generate, I wasn't sure where to put my text. It says in the Output field. With Whitspace, for example, I only have the entire assembler output in a debug mode. I would omit / not display.
We also omitted the debug switch for other things.

t-m-z:
> Is really a lot of expense.
> With Generate, I wasn't sure where to put my text. It says in the Output field. With Whitspace, for example, I only have the entire assembler output in a debug mode. I would omit / not display.
> We also omitted the debug switch for other things.

Yep - however, it was usefull while discussing the listing and find out the correct parts of the poem.
I believe working indoor with the web or with a tablet, the frontend ist suitable. 
I do not think that outside in the woods there is a need for decoding a Beatnik poem.

PR #293, opened 2020-12-27
Feature Cistercian-numbers

t-m-z:
for review
- I am not sure to do the re-using of the segment code in the right way

Additionally
- Due to your advice ragarding open_file, there two further changes in pubspec.yaml and  lib/widgets/utils/file_utils.dart 
- assets/symbol_tables/ogham/AE.png should be changed - I do not know why


t-m-z:
Thanks for pre-check and first hints - To Do
- check indents in widget segment_display and roll back where necessary
- check indents in logic: segment_display and roll back where necessary
- remove unnecessary files (pubspec, etc.)

PR #305, opened 2021-01-05
Vigenere breaker

Mike-3:


assets/i18n/de.json
```diff
@@ -2666,6 +2666,17 @@
   "whitespace_language_start" : "Start",
   "whitespace_language_readable_code" : "Lesbarer Code",
 
+  "vigenere_breaker_title" : "Vigenère Codeknacker",
```
S-Man42, line 2669: 1. Please keep the alphabetical order :) vigenere before whitespace
2. Duplicate keys (three times title, description, example, ...)

.gitignore
```diff
@@ -87,4 +87,14 @@ web
 *.lot
 *.gz
 *.toc
-howtocode/
\ No newline at end of file
+howtocode/
+lib/logic/tools/crypto_and_encodings/general_codebreakers/vigenere_breaker/bigrams/german_bigramsOrg.dart
```
S-Man42, line 91: this exclude is something for your local work, I guess. Maybe you should add a local .gitignore file instead of using the global one.

assets/i18n/en.json
```diff
@@ -2667,6 +2667,17 @@
   "whitespace_language_start" : "Start",
   "whitespace_language_readable_code" : "Readable Code",
 
+  "vigenere_breaker_title" : "Vigenère Code Breaker",
+  "vigenere_breaker_description" : "Chance to automatically break polyalphabetic ciphers",
+  "vigenere_breaker_example" : "ABC with key DEF → DFH",
+
+  "vigenere_breaker_title" : "Vigenère Code Breaker",
```
S-Man42, line 2674: same as DE

lib/logic/tools/crypto_and_encodings/substitution_breaker/guballa.de/breaker.dart
```diff
@@ -0,0 +1,203 @@
+import 'package:gc_wizard/logic/tools/crypto_and_encodings/substitution_breaker/guballa.de/Key.dart';
```
S-Man42, line 1: Here you add the substitution breaker to the old location. Please don't add this to the commit.

lib/logic/tools/crypto_and_encodings/general_codebreakers/vigenere_breaker/vigenere_breaker.dart
```diff
@@ -0,0 +1,114 @@
...
+
+  for (var i = 0; i < keyList.length; ++i)
+    keyList[i].fitnessFiltered = alpha * keyList[i].fitness - (i > 0 ? keyList[i - 1].fitness : keyList[i].fitness) -  (i < keyList.length-1 ? keyList[i + 1].fitness : keyList[i].fitness);
+
+  return keyList;
+}
+
+
+List<List<int>> _createVigenereEncodeSquare(int size, bool beaufortVariant){
+  var vigenereSquare = List<List<int>>(size);
```
S-Man42, line 99: this syntax is decprecated

you can use something like 
`List<List<int>> x = List.generate(size, (index) => <int>[]);`

lib/logic/tools/crypto_and_encodings/general_codebreakers/vigenere_breaker/vigenere_breaker.dart
```diff
@@ -0,0 +1,114 @@
...
+    return VigenereBreakerResult(errorCode: VigenereBreakerErrorCode.OK);
+
+  if (((keyLengthMin < 3) || (keyLengthMin > 1000)) || ((keyLengthMax < 3) || (keyLengthMax > 1000)))
+    // key length not in the valid range 3..1000
+    return VigenereBreakerResult(errorCode: VigenereBreakerErrorCode.KEY_LENGTH);
+
+  var bigrams = getBigrams(alphabet);
+
+  var vigenereSquare = _createVigenereEncodeSquare(bigrams.alphabet.length, vigenereBreakerType == VigenereBreakerType.BEAUFORT);
+  var cipher_bin = List<int>();
```
S-Man42, line 43: this syntax is decprecated. please use something like
```
var cipher_bin = <int>[];
var resultList = <VigenereBreakerResult>[];
```

lib/widgets/tools/crypto_and_encodings/vigenere_breaker.dart
```diff
@@ -0,0 +1,208 @@
+import 'dart:math';
```
S-Man42, line 1: You have to add the vigenere_breaker to 
- Registry
- Main_View
- General Code Breakers Selection

currently I cannot run it

S-Man42: PLease move the widget into subdirectory "general_code_breakers"

S-Man42 (commented):
Worked well for 
GC5WVPF

lib/widgets/tools/crypto_and_encodings/vigenere_breaker.dart
```diff
@@ -0,0 +1,208 @@
...
+
+    return GCWMultipleOutput(
+      children: [
+        _currentOutput.plaintext,
+        Column(
+          children: [
+            GCWOutput(
+              title: i18n(context, 'common_key'),
+              child: GCWOutputText(
+                text: _currentOutput.key
```
S-Man42, line 184: The plain text is in capitel letters. For consistency I'd prefer to output the key in capital letters as well :)

lib/widgets/tools/crypto_and_encodings/vigenere_breaker.dart
```diff
@@ -0,0 +1,208 @@
...
+        Row(
+          children: <Widget>[
+            Expanded(
+              child: GCWText(
+                  text: i18n(context, 'vigenere_breaker_key_length_min') + ':'
+              ),
+              flex: 1
+            ),
+            Expanded(
+              child: GCWIntegerSpinner(
```
S-Man42, line 107: 1. IntegerSpinner have title attributes. So please use them instead of creating your own title with Rows
2. Just an idea: Instead of writing "Key length min" and "Key length max", you could add a GCWDivider(text: "Key Length") and below the two spinners just labeled with "min", "max". I guess this could be more readable, especially in German ("Schlüssellänge" is a long nasty word...)

lib/widgets/tools/crypto_and_encodings/vigenere_breaker.dart
Mike-3: Not realy, but it's ok.

lib/widgets/tools/crypto_and_encodings/vigenere_breaker.dart
Mike-3: I am changing. I had done it that way for reasons of space.

PR #306, opened 2021-01-05
Enhance GCWTool

t-m-z:
**Situation**
Currently there are two different ways to add one single button in the AppBar-Widget:
1.  provide a key xxx_titleTrailing in the json-files and an IconButton-Widget
2.  provide a key _xxx_onlinehel_p in the json-files
To be more flexible and provide more than one button, the GCWTool-Widget has to be enhanced or a second, improved Widget has to be used.

**Discussion**
While initializing the GCWizard-App, the registry.dart builds al toolList = List of GCWTool.
Hence there is only one kind of widget possible.

**Recommendation**
Enhance the GCWTool-Widget to be more flexible. Thus
- provide a new field buttonList
- provide a class to initialize the buttonList in the registry
- provide a method to build the buttonList
This enhancement works with the already working keys for
- Online-Help for FormulaSolver
- Online-Help for Variable Coordinates
- download of SymbolTables-PDF
Hence there are no changes necessary - however could be done in future.

**Example**
```
GCWTool(
     tool: Chef(),
     i18nPrefix: 'chef',
     searchStrings: 'chef koch rezept recipe language programming sprache esoteric esoterisch programmiersprache',
     buttonList: [
         GCWToolActionButtonsEntry(true, 'chef_download_documentation_url', 'chef_download_documentation_title', 'chef_download_documentation_text', Icons.file_download),
         GCWToolActionButtonsEntry(false, 'chef_online_help_url', '', '', Icons.help)],
),
```

### DIFF
--- a/lib/widgets/common/gcw_tool.dart
+++ b/lib/widgets/common/gcw_tool.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/material.dart';
 import 'package:gc_wizard/i18n/app_localizations.dart';
 import 'package:gc_wizard/theme/theme.dart';
+import 'package:gc_wizard/widgets/common/base/gcw_dialog.dart';
 import 'package:gc_wizard/widgets/common/gcw_symbol_container.dart';
 import 'package:gc_wizard/widgets/selector_lists/gcw_selection.dart';
 import 'package:gc_wizard/widgets/utils/common_widget_utils.dart';
@@ -16,6 +17,7 @@ class GCWTool extends StatefulWidget {
   final autoScroll;
   final iconPath;
   final String searchStrings;
+  final List buttonList;
 
   var icon;
   var _id = '';
@@ -36,7 +38,8 @@ class GCWTool extends StatefulWidget {
     this.autoScroll: true,
     this.iconPath,
     this.searchStrings: '',
-    this.titleTrailing
+    this.titleTrailing,
+    this.buttonList
   }) : super(key: key) {
     this._id = className(tool) + '_' + (i18nPrefix ?? '');
     this._isFavorite = Prefs.getStringList('favorites').contains('$_id');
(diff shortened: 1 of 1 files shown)


S-Man42:
First test, opened just one tool:
grafik

lib/widgets/common/gcw_tool.dart
```diff
@@ -16,6 +17,7 @@ class GCWTool extends StatefulWidget {
   final autoScroll;
   final iconPath;
   final String searchStrings;
+  final List buttonList;
```
S-Man42, line 20: You should never declare untyped collections! :) You should take List<Widget> instead or similar

lib/widgets/common/gcw_tool.dart
```diff
@@ -69,17 +72,24 @@ class GCWTool extends StatefulWidget {
   _GCWToolState createState() => _GCWToolState();
 }
 
+class GCWToolActionButtonsEntry {
```
S-Man42, line 75: Please offer an example to how do you expect to use this. Should  the buttons be generated in the registry? Or should they created here from the i18n key like "online_help" by an special identifier (which would be the way I prefer)...

lib/widgets/common/gcw_tool.dart
```diff
@@ -102,6 +112,38 @@ class _GCWToolState extends State<GCWTool> {
     );
   }
 
+          _buildButtons() {
```
S-Man42, line 115: indent issue

lib/widgets/common/gcw_tool.dart
```diff
@@ -69,17 +72,24 @@ class GCWTool extends StatefulWidget {
...
   @override
   Widget build(BuildContext context) {
     return Scaffold(
       appBar: AppBar(
         title: Text(widget.toolName),
-        actions: <Widget>[
-          widget.titleTrailing ?? _buildHelpButton()
-        ],
-      ),
+        actions: _buildButtons(),
```
S-Man42, line 92: indent issue

lib/widgets/common/gcw_tool.dart
t-m-z: As I wrote in the PR-Comment 
- your provide actually two different ways to implement only one button; either to start a download or to provide onlinehelp:
  - titletrailing-key and an additional widget with a button like for example for the download of the symbol-tables.pdf
  - onlinehelp-key and _buildHelpButton-Method inside the GCWTool-Widget like for example for the formula-solver
- I suggest build one flexible method for zero, oner or more buttons, thus
  - to define the buttons in the registry
  - to analyse the key and build the appbar with the buttons in the GCWTool-Widget

t-m-z:
will be re-opened for 1.5.0

PR #307, opened 2021-01-05
feature coding/encoding Cistercian numbers with a segment display

t-m-z:
for first review and discussion


S-Man42:
Please add tool to:
- Registry
- Main View

Currently this is not runnable because of missing entry hooks

S-Man42:
758 is not decodable
grafik

S-Man42:
I wouldn't use the "text encryption":
- "adfjkou" is an internal segment encoding, but nothing, a user must know about. It could yield confusions -> Unenlighted users may think, this is a real cistercian encryption. -> Same for decrypt output, which should only accept visual input
- Second: The vertical bar in the middle should always be on, I think.
- So, we need to enhance the segment widget for two points:
-> Add boolean for visualEncryptionOnly
-> Use "notEditable" flag for specific input segments.

lib/logic/tools/science_and_technology/segment_display.dart
```diff
@@ -467,20 +562,19 @@ List<List<String>> encodeSegment(String input, SegmentDisplayType segmentType) {
...
+        if (display != null)
+          output.add(AZToSegment[character]);
     }
   }
 
   return output;
 }
 
-
 Map<String, dynamic> decodeSegment(String input, SegmentDisplayType segmentType) {
```
S-Man42, line 575: To extract the cistercian logic from the segment logic, this could take an additional parameter: 
decodeSegment(String input, SegmentDisplayType segmentType, {customSegments})

lib/logic/tools/science_and_technology/segment_display.dart
```diff
@@ -467,20 +562,19 @@ List<List<String>> encodeSegment(String input, SegmentDisplayType segmentType) {
...
   if (input == null || input == '')
     return {'displays': [], 'text': ''};
-
   var baseSegments;
 
   switch (segmentType) {
```
S-Man42, line 580: You don't even need the SegmentDisplayType.CISTERCIAN. You make it SegmentDisplayType.CUSTOM for reusablitity.

and then:

switch () {
   case SEVEN:...
   case CUSTOM: baseSegments = customSegments // your parameter with your customSegments
}

lib/widgets/tools/crypto_and_encodings/cistercian.dart
```diff
@@ -0,0 +1,283 @@
...
+                  )
+                ],
+              ),
+            ),
+            _buildOutput(countColumns)
+          ]
+      );
+  }
+
+  _buildTextualDecryption(){
```
S-Man42, line 137: I would completely leave it, see above

lib/widgets/tools/crypto_and_encodings/cistercian.dart
```diff
@@ -0,0 +1,283 @@
...
+          return UNKNOWN_ELEMENT;
+
+        return character.join();
+      }).join(' ');
+
+      return Column(
+        children: <Widget>[
+          _buildDigitalOutput(countColumns, segments),
+          GCWDefaultOutput(
+              child: output
```
S-Man42, line 257: See above: I would leave out the textual output.

S-Man42 (commented):
Looks very well so far! You the logic extraction and removing the textual input/output...

Well done!

t-m-z:
class is registered in registry and main_view
cistercian logic is extracted
758 can be decoded
however I have a strange behavoiur with the decoding logic: 6, 60, 600 and 6000 can not be decoded. Their characters d, c, o and n are not recognized. I need your help. I do not see this error.

PR #308, opened 2021-01-05
feature Number sequences

t-m-z:
Allows calculation on number sequences.

Calculations are
- check given number if part of sequence and return position or -1
- get nth number of sequence
- get numbers with n digits
- get numbers within a given range
- check given sequence of digits if part of sequence and return position or -1

Number sequences are
- Lucas
- Fermat
- Mersenne
- Pell
- Jacobsthal
- Pell-Lucas
- Lucas-Jacobsthal
- Recaman
- Catalan


S-Man42:
Created PR onto your branch for optimizing the output widgets, texts and descriptions:
https://github.com/t-m-z/GCWizard/pull/4

lib/logic/tools/science_and_technology/number_sequence.dart
```diff
@@ -0,0 +1,860 @@
...
+}
+
+List getNumbersInRange(NumberSequencesMode sequence, int start, stop) {
+  if (start == null || stop == null || start == '' || stop == '')
+    return [];
+
+  List numberList = new List();
+
+  if (sequence == NumberSequencesMode.FERMAT) {
+    for (int i = start; i <= stop; i++)
```
S-Man42, line 212: Well... a lot of code duplication right in here...

you can avoid a lot, when you do this:

```
Function _getNumberSequenceFunction(NumberSequencesMode mode) {

  switch (mode) {
    case NumberSequencesMode.FERMAT: return _getFermat;
    case NumberSequencesMode.MERSENNE: return _getMersenne;
    case NumberSequencesMode.CATALAN: return _getCatalan;
    case NumberSequencesMode.JACOBSTAHL: return _getJacobsthal;
    case NumberSequencesMode.JACOBSTHAL_LUCAS: return _getJacobsthalLucas;
  }

  return null;
}
```
and then use this like this:

```
var numberSequenceFunction = _getNumberSequenceFunction(sequence);
if (numberSequenceFunction != null) {
	for (int i = start; i <= stop; i++) {
	  numberList.add(numberSequenceFunction(i)); // use function here
	}
} else if (sequence == NumberSequencesMode.FIBONACCI) {
...
}...
```

Same can be done on the other functions:

```
    index = 0;
    number = numberSequenceFunction(index);  // here
    while (number <= checkNumber){
      if (number == checkNumber) {
        return index.toString();
      } else
        index++;
      number = numberSequenceFunction(index); // here
    }
```

S-Man42: This avoid the code duplication for at least 5 functions.

lib/logic/tools/science_and_technology/number_sequence.dart
```diff
@@ -0,0 +1,860 @@
...
+  else if (sequence == NumberSequencesMode.JACOBSTAHL) {
+    for (int i = start; i <= stop; i++)
+      numberList.add(getJacobsthal(i).toString());
+  }
+  else if (sequence == NumberSequencesMode.JACOBSTHALLUCAS) {
+    for (int i = start; i <= stop; i++)
+      numberList.add(getJacobsthalLucas(i).toString());
+  }
+  else if (sequence == NumberSequencesMode.FIBONACCI) {
+    BigInt number;
```
S-Man42, line 232: Then you are writing the Fibonacci (and other) code again and again.

This can be avoided for example:
- Search for n is the same as search for a range from n to n. So you can reuse the range function for checkAt. checkAt will return an one-elemented array

- Check for digits and and check for a number is also the same if you convert the number to digits...

So you could write a function like this (pseudo):

```
PositionSequenceOutput checkDigits(number, index, digitsToCheck) {
    if (digitsToCheck == null)
      return null;

    var idx = number.toString().indexOf(digitsToCheck);
    if (idx >= 0)
        return PositionSequenceOutput(number, index, idx + 1);

    return null;
}

dynamic doFibonacci (int start, int end, {String digitsToCheck, int countDigits}) {
   var number;
    while (index <= end) {
       number = index + previousNumber // or however the iteration step looks like
       
       if (index >= start) {   // relevant range
           var x = checkDigits(number, index, digitsToCheck);   // for checkNumber and checkDigits
           if (x != null)
                return x;

           var cnt = countDigits(number, countDigits);          // for countDigits
           if (cnt != null && cnt == true) {
               numbers.add(number);
           }
           if (cnt.toString().length > countDigits)
              return numbers;

           if (start < end || index == end)           // for range, second condition meets if for getNth
               numbers.add(number);

       }  

      index++; 
   }

   return numbers;
}


getNthFibonacci(n) {
    return doFibonacci(n, n)[0];
}

getFibRange(a, b) {
    return doFib(a, b)
}

getCheckNumber(a) {
    var x = doFib(0, maxE, digitsToCheck: a.toString())
    if (x.positionNumber == a)
       return true;
}

getCheckDigits(a) {
    return doFib(0, maxE, digitsToCheck: a.toString())
}

getNumbersWithDigits(a) {
   return doFib(0, maxE, countDigits: a);
}
```

Moreover: You can optimize this more... The part where you calc the next iteration can be replaced by a function like this, i guess:

```
calcNextFib(a0, b0) {
     return a0 + b0
}
```

This enables you to use the function even more general.

instead of `doFibonacci()` you can make it 

```
doIterationFunction(...., SequenceMode mode) {
    var iterateStep = getFunctionForMode (mode); // as show for the first five functions    

    while (index < end){ 
         number = iterateStep();
         
         if (index >= start) {
             // same code as above
         }
   }
}
```

lib/logic/tools/science_and_technology/number_sequence.dart
```diff
@@ -0,0 +1,860 @@
...
+// https://oeis.org/A084175   Jacobsthal-Oblong   a(n) = a(n-1) * a(n-2)
+// https://oeis.org/A008336   RecamánII          	a(n+1) = a(n)/n if n|a(n) else a(n)*n, a(1) = 1.
+// https://oeis.org/A000058   Sylvester           a(n) = 1 + a(0)*a(1)*...*a(n-1)
+// https://oeis.org/A000110   Bell                a(n) = exp(exp(n) - 1)
+
+
+import 'dart:math';
+
+class getPositionOfSequenceOutput {
+  final String number;
```
S-Man42, line 27: It is not clear to me, why you are using String output instead of (BigInt, int, int). You always cast your results into Strings...

The String is just for the output. It's a frontend type in this case. I'd prefer to convert the very final result once in the output.

lib/logic/tools/science_and_technology/number_sequence.dart
```diff
@@ -0,0 +1,860 @@
...
+final Zero = BigInt.zero;
+final One = BigInt.one;
+final Two = BigInt.two;
+final Three = BigInt.from(3);
+final sqrt5 = sqrt(5);
+final sqrt2 = sqrt(2);
+
+
+BigInt getFermat(int n){
+  return Two.pow(n) + One;
```
S-Man42, line 60: According to https://en.wikipedia.org/wiki/Fermat_number the formula ist not (2^n - 1) but (2^2^n - 1), which let the function explode! :D

S-Man42 (commented):
A lot of stuff here to optimize, especially the code duplication in the logic. I would like to see a more optimized code there. If you don't have any time, I would merge it nonetheless but would create a "technical dept" ticket for the next release. However, the wrong calculation of Fermat must be fixed in any way.

Additional points:
- Please think about adding a submit button. On some function, when you are typing a 4 digit input, the calculation takes much time. In that time, entering a 5th digit or a 6th, is not possible. The better approach here is to let the user input the whole input and then run the calculation.
- A wish for the end: Factorial as separate sequence would be quite nice, and is not that uncommon to use, I guess.

t-m-z:
a lot of food for thougth - but valid.
- searchstrings are corrected
- code-dublication reduced
- cast to string ist part of the frontend
- Fermat formula corrected
- Factorial added
- Fermat renamed to Mersenne-Fermat
- description according to OEIS added

t-m-z:
test were added and successful

t-m-z:
ragarding maxIndex, maxDigits: field are added and initialized. The correct values have to explored

S-Man42:
Closed because of other branch merged: #319

PR #324, opened 2021-01-17
add numbersequences

t-m-z:
- mersenn primes,
- mersenne exonents, perfect/superperfect/pseudoperfect/sublime/weird numbers
- jacobsthal-oblong
- bell


S-Man42 (commented):
Please check:
- all n: max n not too high (see Bell, 555 seems the upper threshold, I guess... My emulator doesn't make much more). 
- whether chosen n matches Integer Spinner Max (I guess you have always chosen Min = n - 1. You should set it to n :) )
- ALL examples!!! I didn't check all, but some are wrong (Recaman examples at Mersenne, e.g.)

t-m-z:
Thanks for the checks - sometimes there are to many trees in the forest.
Regarding the ranges - I recommend do check these with the next test version. Crashing the emulator is not an option I believe.

PR #345, opened 2021-02-05
progress bar cancel extension

Mike-3:
Cancellation possible at an earlier point in time


S-Man42:
Already merged into 1.4.1

PR #349, opened 2021-02-10
gcw tool enhanced

t-m-z:
there are two examples: Beatnik and Chef


assets/i18n/de.json
```diff
@@ -426,6 +430,10 @@
 
   "chef_title" : "Chef",
   "chef_description" : "Esoterische Programmiersprache, die wie ein Kochrezept aussieht",
+  "chef_online_help_url" : "https://blog.gcwizard.net/manual/de/whitespace/02-wie-interpretiert-der-gc-wizard-whitespace-programme/",
+  "chef_download_documentation_url" : "https://blog.gcwizard.net/wp-content/uploads/2021/02/chef_de.pdf",
+  "chef_download_documentation_title" : "Herunterladen der Kurzanleitung",
```
S-Man42, line 435: Is this still necessary? AFAIK we decided to wait for the manual with the links and do not add separate ways for these tools.

lib/widgets/common/gcw_tool.dart
```diff
@@ -9,13 +10,35 @@ import 'package:url_launcher/url_launcher.dart';
...
+class GCWToolActionButtonsEntry {
+  final bool showDialog; // true, if the button should provide a dialog
+  final String url; // url for a download or website
+  final String title; // title-string to be shown in the dialog
+  final String text; // message-text to be shown in the dialog
+  final IconData icon; // icon tto be shown in the appbar
+
+  GCWToolActionButtonsEntry(this.showDialog, this.url, this.title, this.text,
+      this.icon);
+/*  Example for usage in the registry.dart
```
S-Man42, line 22: I guess, when we implemented this into the registry, this comment is not required anymore, isn't it?

lib/widgets/common/gcw_tool.dart
```diff
@@ -76,14 +100,49 @@ class _GCWToolState extends State<GCWTool> {
     return Scaffold(
       appBar: AppBar(
         title: Text(widget.toolName),
-        actions: <Widget>[
-          widget.titleTrailing ?? _buildHelpButton()
-        ],
+          actions: _buildButtons(),
+//        actions: <Widget>[
+//          widget.titleTrailing ?? _buildHelpButton()
+//        ],
```
S-Man42, line 106: Pls, remove unnecessary code.

lib/widgets/common/gcw_tool.dart
```diff
@@ -76,14 +100,49 @@ class _GCWToolState extends State<GCWTool> {
...
+  List<Widget>_buildButtons() {
+    List<Widget> buttonList = new List<Widget>();
+
+    if (widget.titleTrailing.toString() != 'null')
+      return [widget.titleTrailing];
+
+    if (widget.buttonList == null)
+      return [_buildHelpButton()];
+
+    widget.buttonList.forEach((button) {
```
S-Man42, line 121: This opens a dialog for any button. Is this expected? (To be honest, I haven't in mind which functions we want to support with the buttons, but do all need a pre-dialog mandatorily?)

assets/i18n/de.json
t-m-z: Hm - I believe we missunderstand each other. Of course this is necessary. 
This is an example how the de.json and en.json have to be adjustes to make the enhanced gcw_tool work.

lib/widgets/common/gcw_tool.dart
t-m-z: It depends. If you believe the implementation does not need any further documentation, this comment is not required anymore.
Otherwise I would suggest to keep a kind of documentation in the source file.

lib/widgets/common/gcw_tool.dart
t-m-z: I kept it to show the difference - but of course it will be removed.

lib/widgets/common/gcw_tool.dart
t-m-z: As far as I understood the code, that will not be happen.

I did this to keep the enhanced gcw_tool still compatible with the current solution. That means 
1. the titleTrailing key is still analyzed. If there is a key in the .json this specific widget for downloading a file will be returned. Otherwise the actionList entry will be analyzed.
2. if there is a actionLis entry in the registry, this entry will be analyzed and the button bar will be created.
3. if there is no actionList entry in the registry, there will be a check, if there is onlineHelpKey in the .json
4. if there is an onlineHelpKey a button will be displayed. Otherwise there will be an empty container

In the end, there should be only the enhanced gcw_tool to be used. Hence we can delete the titleTrailing key and simplify the code.

t-m-z:
Providing the manual is still a lot of work:
- Currently we have written ~514 articles and I guess, we still have to write about 150 more.
- Every article has to be proofread regarding spelling, typography, layout, links etc.
- Some articles have to be linked
- For every tool has at least a link to the manule to be implemented which means additinal work in the de.json, en.json and registry. For some tools it might be useful to provide an offline-documentation also which means to add a download button.

Hence an early decision how to interlink the app and the manual would be helpful to organize the time   ;-)

assets/i18n/de.json
```diff
@@ -426,6 +430,10 @@
 
   "chef_title" : "Chef",
   "chef_description" : "Esoterische Programmiersprache, die wie ein Kochrezept aussieht",
+  "chef_online_help_url" : "https://blog.gcwizard.net/manual/de/whitespace/02-wie-interpretiert-der-gc-wizard-whitespace-programme/",
```
Birdy2404:
```suggestion
  "chef_online_help_url" : "https://blog.gcwizard.net/manual/de/chef/02-wie-interpretiert-der-gc-wizard-ein-chef-programm/",

```

t-m-z:
examples for beatnik and chef were removed - de.json, en.json, registry are corrected

PR #354, opened 2021-02-14
Update registry.dart - Sorting Selector List for Hashes

t-m-z:


S-Man42:
Hm, I don't know... I ordered the hashes this way for commonly usage. I never needed a Blake Hash, but usually MD5 or SHA-X. To avoid endless scrolling, I put the most common variants first. Is that not ok?

S-Man42:
After consultation with @t-m-z closed with "won't fix". IMHO it's better to keep the actual order, since MD5 and SHA-X are commonly used and should appear first, wherea Blake or something not...

PR #366, opened 2021-03-02
Issue open map export button

Mike-3:
Export in json format (file) 
Import from clipboard
GCWCoordsPasteButton changed to GCWPasteButton (gcw_map_view)


S-Man42:
Well saving might work (even if i could find the file in the directories using the emulator, so I was not able to check if loading works...) However, it's not the behaviour I expected.

Use case: I want to share may current map view. So I want to create the JSON and copy it into... WhatsApp or whatever.

Now I need to go to the file, find an app which can open a JSON file, mark the content, copy and put it into WhatsApp...

I believe it would be better to get the option to copy it right from the app. Like I did with the Text View of the Formula Solver Export. There I present the JSON in the text field, added a Copy button and so you are able to share the content directly with your mates. 

So, maybe we can thing about: When you show the "Successful saved" dialog, add the text view (like we are showing the images we stored), put a Copy button below it, ready. Now you can put this into the messenger, where your mate can easily copy and paste it into his GCW.

Mike-3:
ok, I change this.

S-Man42:
Importing issues:
1. Imported data will not be persisted. So if imported and you leave the tool and re-enter it: The data is gone.
2. Imported data will not be added to the current view, but replaces it. So if you already have some points, after inserting the pasted data, the original data is gone. (for example try to insert twice)

S-Man42:
Sorry for this never-ending story :( 

But: I copied the output and then I removed the entire objects, so I got an empty map. Then I inserted from clipboard:

```
D/CompatibilityChangeReporter(21996): Compat change id reported: 147798919; UID 10151; state: ENABLED
E/flutter (21996): [ERROR:flutter/lib/ui/ui_dart_state.cc(186)] Unhandled Exception: Bad state: No element
E/flutter (21996): #0      ListMixin.firstWhere (dart:collection/list.dart:167:5)
E/flutter (21996): #1      MapViewPersistenceAdapter._mapPolylineDAOToGCWMapPolyline.<anonymous closure> (package:gc_wizard/widgets/tools/coords/map_view/mapview_persistence_adapter.dart:64:41)
E/flutter (21996): #2      MappedListIterable.elementAt (dart:_internal/iterable.dart:411:31)
E/flutter (21996): #3      ListIterator.moveNext (dart:_internal/iterable.dart:340:26)
E/flutter (21996): #4      new _GrowableList._ofEfficientLengthIterable (dart:core-patch/growable_array.dart:188:27)
E/flutter (21996): #5      new _GrowableList.of (dart:core-patch/growable_array.dart:150:28)
E/flutter (21996): #6      new List.of (dart:core-patch/array_patch.dart:50:28)
```

I guess you are trying to add some data into a non-existing list or try to fetch the first element from an empty list.

Now you know, why this is so crappy code: The edge cases are really weird... Sorry for this. But I feel, that we come to end!

Maybe one little thing, nothing important and absolute no must-have. But maybe you can check and if it is only a small thing: Because we are not exporting UUIDs anymore, the JSON object could be renamed to "id" or something. What do you mean?

Mike-3:
I actually did too.
Renaming that makes sense. Maybe we can shorten a few names right away.
latitude => lat
longitude => lon
coordinateFormat => format
circleColorSameAsColor => sameColor
..

PR #368, opened 2021-03-12
restructuring search, integrate online help de, en, fr

t-m-z:


S-Man42:
Two issues:

- You removed the empty lines in the i18n files... Was that intended? Tbh, I like the empty line breaks...
- For fr you added empty URLs. So, what is the process behind. Do French people not see a manual, or does this mean, you automatically show en (which I would strictly prefer...)

PR #371, opened 2021-03-13
fixes in languages

ludoo0d0a:
Remove empty string, so that fallback take place
remove a duplicated key
fix a "von" in english
Correct some french translations


ludoo0d0a:
Languares are up-to-date in crowdin

PR #387, opened 2021-03-14
closes #372, #373, #374, #375, #376, #377, #386  Additional Symbols

### LINKED ISSUE #372: Symboltable Hieroglyphs

t-m-z:
- Chain of Death: Direction
- Chain of Death: Pairs
- Covenant
- Doop Speak
- Hieroglyphs
- Mandalorian
- Nazcaan
- Nyctography

All Symbol tables were shown correctly in the emulator


S-Man42:
I merged this, but now I see some issues, please fix them:

- Hieroglyphs: There's a "LOGO" symbol in the table. I guess, this shouldn't be. Moreover: Something like "SCH" or "DSCH" looks very German. Do you think, it would be better to i18n it? Maybe there's an English variant for it? Don't know, something like "SH" and "DJ" or so? Could you please investigate this.
- Nyctography: THE, LETTER, DATE, ... is something that should be translated as well. Here we need a config file for translation
- I cannot find Mandalorian in the symbol table list... It is "Mando'a" there. Is it the correct name. If so, you should a description on it :) Moreover: The search string contains "startrek"... Really really bad mistake :D "starwars"

t-m-z:
Hieroglyphs
- logo.png removed
- sch, dsch, ch are looking like german - but these reflects also the pronounciation. Hieroglpyhs are much more complex because the symbols represents spelling and pronounciation , hence there are different t, d, h, etc.

Nyctography
- THE as well as AND are a kind of short-writing; these should be kept
-  config.file is added

Mandalorian
- the search is corrected - also json-files are checked
- according to the Internet, Mando'a s the name of the language, the description exists already

PR will follow

PR #394, opened 2021-03-16
closes #378 feature: add trivial substitutes

### LINKED ISSUE #378: Brainf**k subsitutes

t-m-z:
implement a switch to generate/interpret trivial substitutes like
ALPHUCK, BINARYFUCK, BLUB, BTJZXGQUARTFRQIFJLV, COLONOSCOPY, COW, DETAILEDFUCK, FLUFFLEPUFF, FUCKBEES, GERMAN, KENNYSPEAK, KONFUCK, MORSEFUCK, NAK, OMAM, OOK, PENISSCRIP, PIKALANG, REVERSEFUCK, REVOLUTION9, ROADRUNNER, SCREAMCODE, TERNARY, TRIPLET, UWU, ZZZ


S-Man42:
Well... ;)

I never used the F**k word in Brainf**k because of some strange US policies... Now you added some derivates which explicitly uses this word... What do you mean? Risk breaking US policies? Try out what may happen? Or systematically avoid this...

t-m-z:
I recognized the ** but not the underlying US-Policies. This is a valid point.
Hence I suggest to  use the correct words for the search but not in the drop-down.
So we stick to the policies but allow the tool to be found with its real name.

Sent from MailDroid

-----Original Message-----

S-Man42:
I even left out the search term at the moment... So you cannot find the tool when you search for "brainfuck"... maybe we could try... I don't know if apple or google reviews would create any alarms...

t-m-z:
we should give it a try. the searchstrings are part of the date so google or apple should not notice it

Sent from MailDroid

-----Original Message-----

t-m-z:
changed the F-word within the logic and the widget

S-Man42:
I believe we already had the discussion. But do you think, that someone who finds a Pokemon Cache would search the Code under "Brainfuck"? I know you love dropdowns for variants... However, I would love to see these codes also as stand-alone tools... However, at least they should get search strings.

assets/i18n/de.json
```diff
@@ -367,12 +367,23 @@
   "brainfk_online_help_url" : "https://blog.gcwizard.net/manual/de/brainfk-und-ook/01-was-ist-die-programmiersprache-brainfk/",
   "brainfk_interpret" : "Interpreter",
   "brainfk_generate" : "Generator",
+  "brainfk_original" : "Original",
+  "brainfk_substitition" : "Substitut",
```
S-Man42, line 371: The better word is "Derivat"

assets/i18n/en.json
```diff
@@ -367,12 +367,23 @@
   "brainfk_online_help_url" : "https://blog.gcwizard.net/manual/en/brainf-and-trivial-substitutions-like-ook/01-what-is-the-esoteric-programming-language-brainfk/",
   "brainfk_interpret" : "Interpreter",
   "brainfk_generate" : "Generator",
+  "brainfk_original" : "Original",
+  "brainfk_substitition" : "Substitut",
```
S-Man42, line 371: Substitut"E" or Derivate

assets/i18n/fr.json
```diff
@@ -321,12 +321,23 @@
   "brainfk_example": ">++++++++[<++++++++>-]<++.",
   "brainfk_interpret": "Interpréteur",
   "brainfk_generate": "Générateur",
+  "brainfk_original" : "Original",
+  "brainfk_substitition" : "Substitut",
```
S-Man42, line 325: I have no idea ;) But I guess here are typos as well... "Original_E_" maybe?

S-Man42:
And maybe remove the "PenisScript" to avoid conflicts with the US...

assets/i18n/de.json
t-m-z: done

assets/i18n/en.json
t-m-z: derivate, done

assets/i18n/fr.json
t-m-z: Original, Dérivée - done

t-m-z:
> And maybe remove the "PenisScript" to avoid conflicts with the US... (Although I find it kinda funny :D )

I changed it to PSSCRIPT and P***sScript for the UI

t-m-z:
> I believe we already had the discussion. But do you think, that someone who finds a Pokemon Cache would search the Code under "Brainfuck"? I know you love dropdowns for variants... However, I would love to see these codes also as stand-alone tools... However, at least they should get search strings.

I see that point - However, there are well known derivates like Okk, which I believe should have an own entry and other less common derivates which would flood the Esoteric language selection. 
Hence I recommend to add searchstrings, but not specific tools.
I belive providing a possibility to customize the interpreter/generator will offer utmost flexibility.

Birdy2404:
How about edding it into the description "incl. Ook, Pokemon, ..." So if you get it as a result by the search string, you can pick the relevant tool also by reading the description.

t-m-z:
> How about edding it into the description "incl. Ook, Pokemon, ..." So if you get it as a result by the search string, you can pick the relevant tool also by reading the description.

Good idea - changed the description in the json-files

S-Man42:
Manually merged due to conflicts:

https://github.com/S-Man42/GCWizard/commit/6771d80f55948a020b5a1955d9456ea64e831aef

PR #398, opened 2021-03-17
Bugfix paste coordinates in mapview

Mike-3:


S-Man42:
What was the problem?

Mike-3:
No coords paste in mapview possible.
Problem caused by chanching the return format from parsing. Problem only in test-version.

PR #402, opened 2021-03-17
closes #332 enhancement: add switch to maya numbers ...

### LINKED ISSUE #332: Mayan numeral system
The Mayan numeral system was used for two purposes: calculating and calendar.
 Due to the solar year, the Mayan calendar uses a variation of the vigesimal system: the second position ends at 17, the third at position ends at 360.  Subsequent positions use all twenty digits and the place values continue as 18×20×20 = 7,200 and 18×20×20×20 = 144,000, etc. See https://en.wikipedia.org/wiki/Maya_numerals

Hence a switch should be incorrporated to differ between normal and calendar numbers.

t-m-z:
to differ between calandar and other calculations

### DIFF
--- a/assets/i18n/de.json
+++ b/assets/i18n/de.json
@@ -1658,7 +1658,7 @@
   "mayanumbers_single_numbers" : "Einzelne Zahlen",
   "mayanumbers_vigesimal" : "Zahl im Vigesimalsystem (Zahlenbasis 20)",
   "mayanumbers_mode_calendar" : "Kalender",
-  "mayanumbers_mode_calculate" : "Sonstig",
+  "mayanumbers_mode_calculate" : "Zahlen",
   "mexicanarmycipherwheel_title": "Mexican Army Chiffrierscheibe",
   "mexicanarmycipherwheel_description": "Verwendet von der Mexican Army im Krieg gegen die USA kurz vor dem 1. Weltkrieg",
   "mexicanarmycipherwheel_example": "TEST → 09580897",
--- a/assets/i18n/de.json
+++ b/assets/i18n/de.json
@@ -1657,6 +1657,8 @@
   "mayanumbers_online_help_url" : "https://blog.gcwizard.net/manual/de/maya/02-wie-werden-maya-zahlen-ent-bzw-verschlusselt/",
   "mayanumbers_single_numbers" : "Einzelne Zahlen",
   "mayanumbers_vigesimal" : "Zahl im Vigesimalsystem (Zahlenbasis 20)",
+  "mayanumbers_mode_calendar" : "Kalender",
+  "mayanumbers_mode_calculate" : "Sonstig",
   "mexicanarmycipherwheel_title": "Mexican Army Chiffrierscheibe",
   "mexicanarmycipherwheel_description": "Verwendet von der Mexican Army im Krieg gegen die USA kurz vor dem 1. Weltkrieg",
   "mexicanarmycipherwheel_example": "TEST → 09580897",
--- a/lib/logic/tools/science_and_technology/numeral_bases.dart
+++ b/lib/logic/tools/science_and_technology/numeral_bases.dart
@@ -2,6 +2,10 @@ import 'dart:math';
 
 const _alphabet = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
 
+List<int> mayaSystem = [1, 20, 360, 7200, 144000, 2880000, 57600000, 1152000000, 23040000000, 460800000000, 9216000000000, 184320000000000,3686400000000000];
+Map <String, int> mayaSystemPosition = {'0' : 0, '1' : 1, '2' : 2, '3' : 3, '4' : 4, '5' : 5, '6' : 6, '7' : 7, '9' : 9,
+  'A' : 10, 'B' : 11, 'C' : 12, 'D' : 13, 'E' : 14, 'F' : 15, 'G' : 16, 'H' : 17, 'I' : 18, 'J' : 19};
+
 //TODO: Refactoring - Still bad style because of bad original Java Code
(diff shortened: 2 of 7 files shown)


S-Man42:
To be honest, I would love to see this as separate tool. We already have a Maya selector, so it could be added as Maya Calendar tool. Moreover, that can be linked into the Date/Calender Section as well :) And this spares the switch option... What do you think?

t-m-z:
Hm – tricky question. If the user could be sure, that the Mayan numbers are uses as an calendar I would agree.
If you do not know whether these symbols ar real vigesimal or calendar-vigesimal a switch would be helpful.

 

Von: MLorenz42 ***@***.***> 
Gesendet: Wednesday, 17 March 2021 22:09
Betreff: Re: [S-Man42/GCWizard] #332 enhancement: add switch to maya numbers ... (#402)

 

To be honest, I would love to see this as separate tool. We already have a Maya selector, so it could be added as Maya Calendar tool. Moreover, that can be linked into the Date/Calender Section as well :) And this spares the switch option... What do you think?

S-Man42:
I guess, in doubt he sees both tools and tries both... Hm... I am really no friend of too much options. And in that case for me it's a quite different tool: Maya Numbers and Maya Calendar

t-m-z:
I understand your point but I do not agree in this case.

The Mayan numbering system was used in two ways:
- doing business calculations
- doing calendarian calculations

For calendarian calculations the system was slightly changed at the second position to get a maximum of 360 days.

Additionally think from a user perspective:
you typed a bunch of symbols and realize, you used the wrong tool and have to retype all the symbols. 
What do you feel.

Sent from MailDroid

-----Original Message-----

t-m-z:
another option would be a two-column output: left column vigesimal system right column calendarian system

Sent from MailDroid

-----Original Message-----

assets/i18n/de.json
```diff
@@ -1657,6 +1657,8 @@
   "mayanumbers_online_help_url" : "https://blog.gcwizard.net/manual/de/maya/02-wie-werden-maya-zahlen-ent-bzw-verschlusselt/",
   "mayanumbers_single_numbers" : "Einzelne Zahlen",
   "mayanumbers_vigesimal" : "Zahl im Vigesimalsystem (Zahlenbasis 20)",
+  "mayanumbers_mode_calendar" : "Kalender",
+  "mayanumbers_mode_calculate" : "Sonstig",
```
S-Man42, line 1661: Typo. Furthermore, I'd rename it to "Numbers" - and change the order. First (default) numbers, right calender as special case.

lib/logic/tools/science_and_technology/numeral_bases.dart
```diff
@@ -274,3 +278,26 @@ String _doubleDecToBase(double num, int base, String alphabet) {
 
   return out;
 }
+
+String convertDecToMayaCalendar(String input) {
```
S-Man42, line 282: Please move this into the maya numbers. This is nothing very common, so it shouldn't be added to the very common numeral_bases. For maya calcs there's already a specific class :)

lib/logic/tools/science_and_technology/numeral_bases.dart
```diff
@@ -2,6 +2,10 @@ import 'dart:math';
 
 const _alphabet = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
 
+List<int> mayaSystem = [1, 20, 360, 7200, 144000, 2880000, 57600000, 1152000000, 23040000000, 460800000000, 9216000000000, 184320000000000,3686400000000000];
+Map <String, int> mayaSystemPosition = {'0' : 0, '1' : 1, '2' : 2, '3' : 3, '4' : 4, '5' : 5, '6' : 6, '7' : 7, '9' : 9,
```
S-Man42, line 6: IMO, this is not necessary. The _alphabet is already what you need. with convertBase(input, 10, 20) you can convert the base 20 system there and back. (have a look at my maka code... it's a one-liner with only using the already existing tools)

assets/i18n/de.json
t-m-z: done

lib/logic/tools/science_and_technology/numeral_bases.dart
t-m-z: basicalle right - but the maya calendar system is slightly different: 
it is not 1, 20, 400, 8000 and so on 
but 1, 20, 360, 720, and so on
Hence convertBase does not work.

lib/logic/tools/science_and_technology/numeral_bases.dart
t-m-z: done

t-m-z:
Thanks for your hints, I would still stick to the option switch beacuse these are two kinds of interpreting numeral symbols.
It does not belong to a calendar tool to convert between different systems. Maybe this could be an issue for future work
https://en.wikipedia.org/wiki/List_of_calendars

Regarding a single Mayan number tool maybe there should be an enhancement providing data from
https://www.indianerwww.de/indian/maya-schrift-kalender.htm#:~:text=Dazu%20benutzten%20sie%20das%20Zw%C3%B6lfersystem%2C%20mit%20den%20auch,20%20kin.%2020%20Tage%20bilden%20einen%20Monat%20uinal. or other esoteric web sites.

t-m-z:
I will close this PR and reopen it as a different tool to calculate Mayan calendars.
Hence the output will be according to the 
https://en.wikipedia.org/wiki/Maya_calendar
a date in 
- Long count https://en.wikipedia.org/wiki/Mesoamerican_Long_Count_calendar
- Haab https://en.wikipedia.org/wiki/Haab%CA%BC
- Tzolkin https://en.wikipedia.org/wiki/Tzolk%CA%BCin

PR #416, opened 2021-03-21
Issue open map import kml/gpx

Mike-3:
restore circles
without file picker
at the moment only paste from clipboard possible
prepared for file import of gpx, kml and kmz


lib/logic/tools/coords/import/gpx_kml_import.dart
```diff
@@ -0,0 +1,303 @@
...
+      lines.removeAt(i);
+    }
+  }
+}
+
+bool _isClosedLine(GCWMapPolyline line) {
+  return  ((line.points.first.point.latitude - line.points.last.point.latitude) < epsilon) && ((line.points.first.point.longitude - line.points.last.point.longitude) < epsilon);
+}
+
+bool _completeCircle(GCWMapPolyline line, List<GCWMapPoint> points) {
```
S-Man42, line 240: Haha... Do you re-calculate the Circle-Polyline to a real Circle? :D Great! :D

lib/widgets/utils/file_utils.dart
```diff
@@ -1,4 +1,5 @@
 import 'dart:io';
+import 'dart:typed_data';
```
S-Man42, line 2: What is this for? :) Just curious...

Mike-3: For zip-files (.kmz). Prepares for file picker.

lib/logic/tools/coords/import/gpx_kml_import.dart
S-Man42: One thought which came to my mind here.

I am not quite sure, but AFAIS you are taking every polyline, take 4 points and calculate the middle and recreate a circle from it, correct?

But how can you be sure that the polyline is not a square, for example or a rectangle... How do you distinct a polygon from a circle?

My idea:
1. A circle must have at least... I don't know... 20 (? dont know, maybe 50 or 100??) points. Otherwise, it's still a simple polygone. And, of course, the last point must equal the first point.
2. After calculating the center and radius point, you have to check, that ALL points of the polyline have the same distance to avoid some strange figures... like... a "Kreisbogen" which is not closed, but only, 300 degrees or so and in the end there are some corners... whatever.

I really love your approach, but, AFAIS it handles no edge cases... I know, you would never catch all of them with a heuristic, but with these two simple enhancements, you can catch a lot, I guess... What do you think?

Mike-3: At the moment  I make point 2. Min point count is 4.
1. ckeck Polyline is closed.
2. every distance from the points is the same (2% toller)

Ok. when we have only 4 points, then we have a square. The distance is the same. I must ingrees the min count.

S-Man42:
Please create a PR on 1.6.0

Issue #388, opened 2021-03-14
Buggy Help Links


S-Man42:
* coords_selection_online_help_url is empty in EN (i guess there are some more...)
* Symboltables help link points to PDF download, not to help EN

t-m-z:
regarding "coords_selection_online_help_url is empty in EN (i guess there are some more...)"
online_help_url checked and missing urls are added

regarding "Symboltables help link points to PDF download, not to help EN"
I do not understand because all links seems to be okay:
- "symboltables_selection_online_help_url" : "https://blog.gcwizard.net/manual/en/symboltables/01-which-symbol-tables-are-available/",
- "symboltables_selection_download_link" : "https://misc.gcwizard.net/symboltables_en.pdf",
- "symboltables_selection_online_help_url" : "https://blog.gcwizard.net/manual/de/symboltabellen/01-welche-symboltabellen-stehen-zur-verfugung/",
- "symboltables_selection_download_link" : "https://misc.gcwizard.net/symboltables_de.pdf",

PR #423, opened 2021-03-24
enhance gcw tool with search

t-m-z:
first idea for review and discussion


t-m-z:
closes also #388

S-Man42:
Please fix the registry conflicts.

lib/widgets/common/gcw_tool.dart
```diff
@@ -18,6 +19,21 @@ enum ToolCategory {
   SYMBOL_TABLES
 }
 
+final SearchBlackList = {
+  'der',
```
S-Man42, line 23: ein, eine, un, une, d', l', le, la, of,

t-m-z:
merge problems solved
registry checked
blacklist enhanced

lib/widgets/common/gcw_tool.dart
```diff
@@ -18,6 +19,28 @@ enum ToolCategory {
...
+  'three',
+  'of',
+  'from',
+  'and',
+  'deux',
+  'trois',
+  'de',
+  "l'",
+  'le',
+  'la'
```
S-Man42, line 41: Was it intended that you left out: ein, eine, du (== french de + le), un , une, a (== french à and english a)? One can add "one" as well...

lib/widgets/common/gcw_tool.dart
```diff
@@ -100,19 +126,47 @@ class _GCWToolState extends State<GCWTool> {
...
-
     String url = '';
 
+    // add button with url for searching knowledge base with toolName
+    final Locale appLocale = Localizations.localeOf(context);
+    if (isLocaleSupported(appLocale))
+      url = 'https://blog.gcwizard.net/manual/' +
+          Localizations.localeOf(context).toString() +
+          '/search/' +
+          widget.toolName;
```
S-Man42, line 137: I guess you need to escape this, correct? Or does "Affine Cipher" with spaces work?

lib/widgets/common/gcw_tool.dart
```diff
@@ -100,19 +126,47 @@ class _GCWToolState extends State<GCWTool> {
...
+    else // fallback to en if unsupported locale
+      url = 'https://blog.gcwizard.net/manual/' +
+          'en' +
+          '/search/' +
+          widget.toolName;
+    SearchBlackList.forEach((element) {
```
S-Man42, line 143: It shouldn't be case sensitive... Moreover in cases of french d' and l', there's no space following.

lib/widgets/common/gcw_tool.dart
t-m-z: "Affine cipher" with spaces work

lib/widgets/common/gcw_tool.dart
t-m-z: the blacklist is a living document - so this was the first rough guess

lib/widgets/common/gcw_tool.dart
t-m-z: Because we do a full text search the removals of Diacritics should be avoided

t-m-z:
unfortunately I cannot test the changes due to an error message I do not understand:
A problem was found with the configuration of task ':app:createDebugCompatibleScreenManifests' (type 'CompatibleScreensManifest').
File 'D:\Programmierung\GCWizard\build\app\intermediates\apk_list\debug\apk-list.gson' specified for property 'apkList' does not exist.

I saw an update of gradle to 6.3.8 and no this occurs.
I have already deleted the .gradle but no improvement.

PR #431, opened 2021-03-28
 bugix json and gcw-tool to resolve search problems

t-m-z:


lib/widgets/common/gcw_tool.dart
```diff
@@ -151,14 +162,25 @@ class _GCWToolState extends State<GCWTool> {
     if (text == null) return '';
 
     text = text.trim().toLowerCase();
-    text = text.replaceAll(RegExp(r"[\s+'`´]"), ' ');
-    text = text.split(' ').where((word) => !_SEARCH_BLACKLIST.contains(word)).join(' ');
+    text = text
+        .replaceAll(RegExp(r"[\s+'`´]"), ' ')
+        .replaceAll('d ni', "d'ni")
```
S-Man42, line 167: We should have a _SEARCH_WHITELIST here:
{
'd ni': 'd'ni', ...
}

And then: 

text = text.replace(...)
text = text.split(...).join(' ')
text = substitution(text, _SEARCH_WHITELIST)

lib/widgets/common/gcw_tool.dart
```diff
@@ -151,14 +162,25 @@ class _GCWToolState extends State<GCWTool> {
...
+        .replaceAll('d or', "d'or")
+        .replaceAll('mando a', "mando'a")
+        .replaceAll('**', '')
+        .replaceAll('/', ' ')
+        .replaceAll(RegExp(r"\([a-zA-Z0-9\s.]+\)"), ''); //remove e.g. (128 bits) in hashes-toolname
```
S-Man42, line 172: I already thought about that, but we have titles with dashes: Geocache-Code, Gray-Code, SHA-1, ... So I prefer to keep the dash also... What do you mean?

lib/widgets/common/gcw_tool.dart
```diff
@@ -151,14 +162,25 @@ class _GCWToolState extends State<GCWTool> {
...
+        .where((word) => !_SEARCH_BLACKLIST.contains(word))
+        .join(' ');
     return text;
   }
 
   bool _needsDefaultHelp(Locale appLocale) {
     return !isLocaleSupported(appLocale) ||
-        (widget.missingHelpLocales != null && widget.missingHelpLocales.contains(appLocale.languageCode));
+        (widget.missingHelpLocales != null &&
+            widget.missingHelpLocales.contains(appLocale.languageCode));
```
S-Man42, line 183: Did you execute the formatter? I'd wonder why it formats the code that way for me and the other way for you...

`dart format lib -l 120`

lib/widgets/common/gcw_tool.dart
```diff
@@ -151,14 +162,25 @@ class _GCWToolState extends State<GCWTool> {
     if (text == null) return '';
 
     text = text.trim().toLowerCase();
-    text = text.replaceAll(RegExp(r"[\s+'`´]"), ' ');
-    text = text.split(' ').where((word) => !_SEARCH_BLACKLIST.contains(word)).join(' ');
+    text = text
+        .replaceAll(RegExp(r"[\s+'`´]"), ' ')
```
S-Man42, line 166: Please do me a favor and fix my own code here, I made a mistake (to avoid any conflicts I won't do it by myself):

Replace

`text = text.replaceAll(RegExp(r"[\s+'`´]"), ' ');`

with

```
text = text.replaceAll(RegExp(r"['`´]"), ' ');
text = text.replaceAll(RegExp(r'\s+'), ' ');
```

lib/widgets/common/gcw_tool.dart
t-m-z: yeah - perfect idea. Done.

lib/widgets/common/gcw_tool.dart
t-m-z: I would keep the dash - I did not find any problems with searchstring.
However there are some problems with slash and brackets

lib/widgets/common/gcw_tool.dart
t-m-z: done

PR #434, opened 2021-03-30
add description and searchstrings for new symboltables dorabella unitology al bhed

t-m-z:


t-m-z:
regarding albhed:
- add config file
- add smallcaps in zip-file
- rename zip-file

PR #439, opened 2021-04-01
feature silver ratio

t-m-z:
I tried to get familiar with your way of providing tools for irrational numbers
Maybe we should think about a selection


S-Man42:
Please create a PR on 1.6.0

PR #472, opened 2021-04-11
Tmz 1.5.0 registry missing locales

t-m-z:
change      missingHelpLocales: [] 
        to      missingHelpLocales: ['cs', 'da', 'es', 'it', 'ja', 'ko', 'nl', 'pl', 'ru', 'zh' ]


S-Man42:
Not necessary here, because the list is only relevant for supported locales. None of them are supported at the moment, so we don't need to fill the long lists we irrelevant values

PR #476, opened 2021-04-12
Issue coordinate selector bugs

Mike-3:
big refactoring
base coordinate class
for every format a seperate class (with key, toLatLng, toString)


S-Man42:
Please create a PR on 1.6.0

PR #502, opened 2021-04-26
reactivate open_file (open_file added gpx to supported extension)

Mike-3:
update pubspec.yaml


S-Man42:
Does it work now?

Mike-3:
I have the new verion tested, it works.

PR #506, opened 2021-04-27
Feature symbol tables braille de.en.fr

t-m-z:
every language has specific additions to the original braille like short writing, abbreviations etc.


assets/i18n/en.json
```diff
@@ -2424,10 +2424,20 @@
   "symboltables_blox_description": "Alternative Semitic language",
   "symboltables_brahmi_numerals_title": "Brahmi",
   "symboltables_brahmi_numerals_description": "Indian numerals from the 3rd century BC",
-  "symboltables_braille_title": "Braille",
-  "symboltables_braille_description": "Tactile signs for blind people",
-  "symboltables_braille_capital_follows": "CAPITAL FOLLOWS",
-  "symboltables_braille_number_follows": "NUMBER FOLLOWS",
+  "symboltables_braille_de_title": "Braille, de",
```
S-Man42, line 2427: I'd write "Braille (German)"

assets/symbol_tables/braille_de/config.file
```diff
@@ -0,0 +1,44 @@
+{
+	"case_sensitive" : true,
+	"special_mappings" : {
+		"article" : "Art.",
+		"A" : "a",
+		"AE" : "ä",
```
S-Man42, line 6: This is not clear to me. If it's case sensitive, why do you map all uppercases to lowercases?

t-m-z:
Because the standard of braille is lower acse – or I did not understand the meaning of this switch “caseSensitive”

 

Von: MLorenz42 ***@***.***> 
Gesendet: Sunday, 2 May 2021 17:39
Betreff: Re: [S-Man42/GCWizard] Feature symbol tables braille de.en.fr (#506)

 

@S-Man42 commented on this pull request.

  _____  

In assets/symbol_tables/braille_de/config.file <https://github.com/S-Man42/GCWizard/pull/506#discussion_r624713716> :

> @@ -0,0 +1,44 @@
 
+{
 
+       "case_sensitive" : true,
 
+       "special_mappings" : {
 
+              "article" : "Art.",
 
+              "A" : "a",
 
+              "AE" : "ä",
 

This is not clear to me. If it's case sensitive, why do you map all uppercases to lowercases?

S-Man42:
Well, the default is writing everything in Capital letters, no matter how they are spelled correctly. It's like a crypto decoding: All letters capital.

If you define the symbols case_sensitive, thrn you cannot encode  A and a the same. It would only generate an output for one.

So, if theres no difference for the symbols A and a, please stay case insensitive. Yes, then everything will be uppercase but that shouldnt be a real problem here...

assets/symbol_tables/braille_fr/config.file
```diff
@@ -0,0 +1,24 @@
...
+		"E_circumflex" : "Ê",
+		"E_grave" : "È",
+		"E_trema" : "Ë",
+		"I_circumflex" : "Î",
+		"I_trema" : "Ï",
+		"O_circumflex" : "Ô",
+		"OE_ligature" : "Œ",
+		"U_circumflex" : "Û",
+		"U_grave" : "Ù",
+		"U_trema" : "Ü",
```
S-Man42, line 17: maybe something for the global list in the logic file, what do you think? (in that case I'd rename mult to times)

t-m-z:
thanks for the hint. I will adapt symbol_table_data.dart.


Sent from MailDroid

-----Original Message-----

assets/symbol_tables/braille_fr/config.file
t-m-z: done - it is not always easy to understand the logic with the first look

assets/symbol_tables/braille_en/config.file
```diff
@@ -0,0 +1,8 @@
+{
+	"case_sensitive" : false,
+	"special_mappings" : {
+		"divide" : "÷",
```
S-Man42, line 4: I have a strange encoding here. My notepad++ is showing some Thai encoding instead of UTF-8 which seems to cause some problems. Furthermore: There are already special characters defined for these chars: "devision" and "times". You don't need to add them separately

assets/symbol_tables/braille_de/config.file
```diff
@@ -0,0 +1,7 @@
+{
+	"case_sensitive" : false,
+	"special_mappings" : {
+		"article" : "Art.",
```
S-Man42, line 4: comma is too much which makes it not well-formed JSON

assets/symbol_tables/braille_en/config.file
```diff
@@ -0,0 +1,8 @@
...
+		"mult" : "×",
```
S-Man42, line 5: Please remove the trailing comma from the very last element. Not well-formed JSON ;)

S-Man42:
looks good otherwise :)

assets/symbol_tables/braille_en/config.file
t-m-z: encoding seems to be OK. special_mapping removed

assets/symbol_tables/braille_de/config.file
t-m-z: done

assets/symbol_tables/braille_en/config.file
t-m-z: done - but I had no crash

t-m-z:
PR done

S-Man42:
Did you change the file names as well. If you removed the "divide" and "mult" names from the config files, you need to change the files according to the existing names "division" and "times".

t-m-z:
done

PR #508, opened 2021-04-28
Issue Vigenere Cipher Autokey encrypt problem

Mike-3:
Disable "Ignore Non-Letters" when AutoKey is enabled.
switches flipped

### DIFF
--- a/lib/widgets/tools/crypto_and_encodings/vigenere.dart
+++ b/lib/widgets/tools/crypto_and_encodings/vigenere.dart
@@ -67,14 +67,6 @@ class VigenereState extends State<Vigenere> {
           },
         ),
         GCWOnOffSwitch(
-          title: i18n(context, 'vigenere_ignorenonletters'),
-          onChanged: (value) {
-            setState(() {
-              _currentNonLetters = value;
-            });
-          },
-        ),
-        GCWOnOffSwitch(
           title: i18n(context, 'vigenere_autokey'),
           onChanged: (value) {
             setState(() {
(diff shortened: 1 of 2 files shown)


S-Man42:
Hm, I am not quite sure if that is the correct way to solve this. AFAIK there is logic which handles IgnoreNonLetters and Autokey (please see the tests). So if there's a problem, I guess, this needs to be fixed in the logic not by disabling some options.

Please can you create a test with your use case and check where is the real problem?

Mike-3:
Ok, but i have no online tool found, where you can combine this features. And i'm relatively sure that it can't in principle either.
The problem is, with autokey the text as key extension is used. But the key has only alphabet values. After a few non alphabet characters is the key shorter as the text. And then you have the problem.

S-Man42:
Hm... maybe you are right...

Mike-3:
The last words are also missing in the test.

        'input' : 'Wie ich sehe hast du das Rätsel gelöst. Du wirst jetzt sicherlich die Koordinaten lesen wollen, keine Sorge, dass wirst du. Aber zunächst ',
        'key': 'beteigeuze',
        'autoKey': true,
        'aValue': 0,
        'expectedOutput' : 'Xmx qil ridm pczl ky dsl Lätkve kprödl. Xq nakbx ixlhv wznpgyoqgr rzh Xohvqtrsxrj wpwrx ebpdse, neafa Jgkjy, hrrm ypjl'

S-Man42:
Hm, you're right. Can you fix this... I currently have not the time to dive into this. Is the test wrong or the function?

Mike-3:
Since this combination is not allowed and the test is faulty (there are missing words in the output) I would remove the test.

PR #511, opened 2021-04-30
Exif reader

ludoo0d0a:
First PR for exif reader

We need to validate : 
- tools's name (exif reader or metadata parser or...)
- i18n strings
- UI
- interaction with other widgets (from @Mike-3 's gif decoder)


ludoo0d0a:
#511

ludoo0d0a:
includes basic scrollbar in list
https://github.com/S-Man42/GCWizard/pull/511/commits/1a5adf3886f1364b6e4d0034c693eb066df68036

Mike-3:
-> interaction with other widgets (from @Mike-3 's gif decoder)

my suggestion is, we add a optional PlatformFile property to the widget call .

class ExifReader extends StatefulWidget {
 ExifReader({Key key, PlatformFile platformFile}) : super(key: key);

like on coordinates.
I do this also in my file/ image widgets.

S-Man42:
We have a conflict here now due to the recently merged PRs. Please resolve (merge the current 1.6.0 into your or make a rebase)

ludoo0d0a:
> We have a conflict here now due to the recently merged PRs. Please resolve (merge the current 1.6.0 into your or make a rebase)

I rebased onto 1.6.0. 
Could you tell me it is ok now ?

S-Man42:
I had a look... It's completely weird. If you have a look at the "changed files" here in this PR (a tab at the top), you can see 53! changed files. Chef, Symboltables, Coordinates, GPX/KML Import... That's definitely too much. So it is really not clear to me, what is your new code and what is a mistake...

S-Man42:
For example: Why is there a commit from @t-m-z  in here?

ludoo0d0a:
Ok
I need a proper rebase on 1.6.0 like registry_perf.

ludoo0d0a:
> I had a look... It's completely weird. If you have a look at the "changed files" here in this PR (a tab at the top), you can see 53! changed files. Chef, Symboltables, Coordinates, GPX/KML Import... That's definitely too much. So it is really not clear to me, what is your new code and what is a mistake... I believe rebasing made this even worse :(
> 
> I could try to cherrypick the relevant commits, but it is hard to find them. As I see it: Create a new branch, copy your changed into it and create a new PR with a clean branch... That would be the fastest and most simple way to come clear here...

I merged 1.6.0 in this PR, seems ok now

ludoo0d0a:
Please add tests for the logic. Especially the GPS Section would be very interesting since there are many edge cases to catch (-+90 als limits, so what happens if deg == 90 and minutes == 1? --> there's a normalize function in the coordinate sections... moreover, I'd tend to convert the coords into LatLon using the appropriate function, which already do such checks...; what havens if minutes >= 60, ...)
The test may serve as a documentation as well, so it can be shown all the exif sections we support, if each is tested.

ludoo0d0a:
Maybe I missed it, but what about some metadata: Original File Size, Image Size (width, height, ...)

ludoo0d0a:
In the end I'd tend to name it "Exif and Metadata Viewer" as it's great idol: http://exif.regex.info/exif.cgi :)

ludoo0d0a:
> Maybe I missed it, but what about some metadata: Original File Size, Image Size (width, height, ...)

File section at top displays basic metadata for file:
- filesize
- name
- path
- extension

In exif section, fields for image properties are numerous, including: 
- ExifImageWidth
- ExifImageLength

ludoo0d0a:
> Please add tests for the logic. Especially the GPS Section would be very interesting since there are many edge cases to catch (-+90 als limits, so what happens if deg == 90 and minutes == 1? --> there's a normalize function in the coordinate sections... moreover, I'd tend to convert the coords into LatLon using the appropriate function, which already do such checks...; what havens if minutes >= 60, ...)
> The test may serve as a documentation as well, so it can be shown all the exif sections we support, if each is tested.

basic unit tests added, but I don't know exactly what to check.

ludoo0d0a:
ImageView integrated.

There are some issues in UI / UX

- save
- menu in gcwimageview asking to open in its own tool.
- label 'output' for all sections using GCWMultipleOutput

lib/logic/tools/coords/data/coordinates.dart
```diff
@@ -86,13 +86,25 @@ String _dmmAndDMSNumberFormat([int precision = 6]) {
...
+String _getCoordinateSignString(int sign, bool isLatitude) {
   var _sign = '';
 
-  if (sign < 0) {
-    _sign = isLatitude ? 'S' : 'W';
+  if (isLatitude) {
+    _sign = (sign > 0) ? 'N' : 'S';
   } else {
-    _sign = isLatitude ? 'N' : 'E';
+    _sign = (sign > 0) ? 'E' : 'W';
```
S-Man42, line 95: What if sign == 0? :)

lib/logic/tools/crypto_and_encodings/exif_reader.dart
```diff
@@ -0,0 +1,163 @@
...
+    default:
+      return tag.printable;
+  }
+}
+
+_ascii2string(List<dynamic> values) {
+  return values.map((e) => String.fromCharCode(e)).join();
+}
+
+// int getExifRotation(Map<String, IfdTag> tags) {
```
S-Man42, line 144: Is commentation intended?

lib/widgets/tools/crypto_and_encodings/exif_reader.dart
```diff
@@ -0,0 +1,206 @@
...
+      children: <Widget>[
+        GCWButton(
+          text: i18n(context, 'open_file'),
+          onPressed: _readFileFromPicker,
+        ),
+        Container(),
+        // GCWButton(
+        //   text: i18n(context, 'paste_image'),
+        //   onPressed: _pasteFileFromclipboard,
+        // ),
```
S-Man42, line 56: Is this WIP or can this be deleted?

test/logic/tools/crypto_and_encodings/exif_reader.dart
```diff
@@ -0,0 +1,45 @@
...
+        'ref': 'E',
+        'isLat': false,
+        'expectedOutput': 6.112192666666666,
+      },
+      // deg == 90 and minutes == 1 ; ok we are above 90
+      {
+        'values': [Ratio(90, 1), Ratio(0, 1), Ratio(1, 1)],
+        'ref': 'N',
+        'isLat': true,
+        'expectedOutput': 90.00027777777778,
```
S-Man42, line 26: And that's exactly a coordinate which is not valid. 90 is maximum. So the input is not valid. In the coordinate section we are flexible enough to handle such wrong input by normalization.

So, before you do the output, you should do:

DEC coordinate = DEC(exif_lat, exif_lon);
LatLon latlon = decToLatLon(coordinate); // here is the normalization.

This creates a much better and more realistic coordinate. For 90.1 it returns 89.9 (and the lon + 180, because you passed the north pole by 0.1 degrees and therefore you are on the other hemiphere)

S-Man42: Oh wait, I just see that you did this already in the completeGPSData() method. In that case, don't worry :) But maybe add a test for the completeGPSData() method, which is the more interesting one for the user, of course :)

I personally believe that we should return both: The raw GPS data (with maybe strange data) and of course our normalized LatLon for the coordinate functions/map view. So, we don't lose the original information. What do you think?

lib/widgets/tools/crypto_and_encodings/exif_reader.dart
```diff
@@ -0,0 +1,206 @@
...
+            //Image.memory(thumbnail.bytes),
+            GCWImageView(imageData: thumbnail)
+          ],
+          //suppressCopyButton: false,
+          trailing: GCWIconButton(
+            iconData: Icons.save,
+            size: IconButtonSize.SMALL,
+            //iconColor: _isNoOutput ? Colors.grey : null,
+            onPressed: () {
+              // _isNoOutput ? null : _exportCoordinates(context, widget.points, widget.polylines);
```
S-Man42, line 137: What's the matter with this? Already WIP?

S-Man42:
First try was opening a JPG image -> results in Exception:
```
I/flutter (13758): No EXIF information found
E/flutter (13758): [ERROR:flutter/lib/ui/ui_dart_state.cc(186)] Unhandled Exception: NoSuchMethodError: The method 'containsKey' was called on null.
E/flutter (13758): Receiver: null
E/flutter (13758): Tried calling: containsKey("JPEGThumbnail")
E/flutter (13758): #0      Object.noSuchMethod (dart:core-patch/object_patch.dart:54:5)
E/flutter (13758): #1      completeThumbnail (package:gc_wizard/logic/tools/crypto_and_encodings/exif_reader.dart:24:12)
E/flutter (13758): #2      _ExifReaderState._readFile (package:gc_wizard/widgets/tools/crypto_and_encodings/exif_reader.dart:92:35)
E/flutter (13758): <asynchronous suspension>
E/flutter (13758): 
```
nature-3082832__480

lib/widgets/tools/crypto_and_encodings/exif_reader.dart
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  Future<void> _pasteFile(data) {
+    PlatformFile _file = data.text;
+    return _readFile(_file);
+  }
+
+  Future<void> _readFileFromPicker() async {
+    List<PlatformFile> files = await openFileExplorer(
+      allowedExtensions: ['jpg', 'jpeg', 'tiff'],
```
S-Man42, line 81: is there a reason for this restriction? I usually need this for any file type: at least PNG, GIF, BMP

S-Man42: For test I removed this restriction and tried: https://blog.gcwizard.net/wp-content/uploads/2020/05/cropped-blog_header-2-1.png

This results in an exception. If I put this into the http://exif.regex.info/exif.cgi and this returns many metadata:

<html><body>
<p style="margin-top:40px; margin-bottom:0"><b><big id="PNG">PNG</big></b></p>

Animation | no
-- | --
Bit Depth | 8
Color Type | RGB with Alpha
Compression | Deflate/Inflate
Filter | Adaptive
Image Size | 1,000 × 363
Interlace | Noninterlaced


</body>
</html>

test/logic/tools/crypto_and_encodings/exif_reader.dart
S-Man42: New Category "Images and Files" was created. Please move the Metadata Viewer into this new section :)

lib/widgets/tools/crypto_and_encodings/exif_reader.dart
ludoo0d0a: was a test, don't know if we need/could to manage clipboard right now ?

lib/logic/tools/coords/data/coordinates.dart
S-Man42: Per Definition I guess, 0/0 is N/E. Only negative is S/W... That's the way I'd handle it

lib/widgets/tools/crypto_and_encodings/exif_reader.dart
S-Man42: Yes, that might be a good Idea, however, then this should be implemented in the GCWImageView to enable all points where an image is displayed.

lib/widgets/tools/crypto_and_encodings/exif_reader.dart
ludoo0d0a: png exif parser may be added (if feasible) in a future version.
I don't see a geocacher writing a mystery with png exif...
Let's see with the current scope (jpg+tiff) for the next release.

After that, I prefer to focus on stegano + hidden text in EOF jpg
more common patterns in geaocaching, ins't it ?

ludoo0d0a: metadata for picture (width,height..) : ok, but we need to define which fields is expected

lib/widgets/tools/crypto_and_encodings/exif_reader.dart
ludoo0d0a: I don't know how to manage properly "save file".
I will see what Mike made in gif tool.

lib/widgets/tools/crypto_and_encodings/exif_reader.dart
ludoo0d0a: > First try was opening a JPG image -> results in Exception:
> 
> ```
> I/flutter (13758): No EXIF information found
> E/flutter (13758): [ERROR:flutter/lib/ui/ui_dart_state.cc(186)] Unhandled Exception: NoSuchMethodError: The method 'containsKey' was called on null.
> E/flutter (13758): Receiver: null
> E/flutter (13758): Tried calling: containsKey("JPEGThumbnail")
> E/flutter (13758): #0      Object.noSuchMethod (dart:core-patch/object_patch.dart:54:5)
> E/flutter (13758): #1      completeThumbnail (package:gc_wizard/logic/tools/crypto_and_encodings/exif_reader.dart:24:12)
> E/flutter (13758): #2      _ExifReaderState._readFile (package:gc_wizard/widgets/tools/crypto_and_encodings/exif_reader.dart:92:35)
> E/flutter (13758): <asynchronous suspension>
> E/flutter (13758): 
> ```
> 
> nature-3082832__480

fixed

S-Man42: Metadata defined? That's easy: As much as possible :D

At least:
file size
image size (width, height),
maybe creating/changing date

ludoo0d0a:
> I even tried a JPG, but same exception...

Raise an issue : https://github.com/bigflood/dartexif/issues/22
I cannot explain

ludoo0d0a:
> > I even tried a JPG, but same exception...
> 
> Raise an issue : https://github.com/bigflood/dartexif/issues/22
> I cannot explain

Owner reports a corrupted file...
Not sure that displayed data are real exif fields.

I extracted some of them with Duncan image plugin. 
I completed the image section with these new attributes.

S-Man42:
I tried to open your test2.jpg but resulted in:
```

D/FilePickerUtils( 6915): Allowed file extensions mimes: [image/jpeg, image/jpeg, image/tiff, image/png, image/x-ms-bmp, image/gif, image/webp]
D/FilePickerDelegate( 6915): Selected type */*
I/FilePickerUtils( 6915): Caching from URI: content://com.android.providers.downloads.documents/document/msf%3A69
D/FilePickerUtils( 6915): File loaded and cached at:/data/user/0/de.sman42.gc_wizard/cache/file_picker/test2.jpg
D/FilePickerDelegate( 6915): File path:[com.mr.flutter.plugin.filepicker.FileInfo@836de85]
E/flutter ( 6915): [ERROR:flutter/lib/ui/ui_dart_state.cc(186)] Unhandled Exception: Null check operator used on a null value
E/flutter ( 6915): #0      JpegData._buildComponentData (package:image/src/formats/jpeg/jpeg_data.dart:678:55)
E/flutter ( 6915): #1      JpegData.read (package:image/src/formats/jpeg/jpeg_data.dart:132:11)
E/flutter ( 6915): #2      JpegDecoder.decodeImage (package:image/src/formats/jpeg_decoder.dart:46:10)
E/flutter ( 6915): #3      decodeImage (package:image/src/formats/formats.dart:92:18)
E/flutter ( 6915): #4      _ExifReaderState.completeImageMetadata (package:gc_wizard/widgets/tools/images_and_files/exif_reader.dart:233:12)
E/flutter ( 6915): <asynchronous suspension>
E/flutter ( 6915): #5      _ExifReaderState._readFile (package:gc_wizard/widgets/tools/images_and_files/exif_reader.dart:91:14)
E/flutter ( 6915): <asynchronous suspension>
E/flutter ( 6915): 
```

S-Man42:
IMG_4692.jpg showed data for thumbnail but didn't show the thumbnail itself. Correct?

S-Man42:
DSCN0010.jpg showed no GPS data (only 0/0, 0/0, 0/0, is this correct?)

S-Man42:
test3.jpg: No GPS data

S-Man42:
Recursive Thumb is great! :D :D :D

S-Man42:
Merged it nonetheless to reduce possibility of conflicts. However please have a look at the issues :) Great work nonetheless!

PR #515, opened 2021-04-30
Tmz 1.6.0 divisor of a number

t-m-z:
calculates the divisors of a number


Mike-3 (commented):
grafik

Count Single Digits = 1 ??

t-m-z:
crosstoralwidget ist refactored

t-m-z:
> grafik
> 
> Now Count Length = 1 ;-)

Seems to be a bug in crosstotals.dart. The function is:
int countDigits(String text) {
  if (text == null) return 0;
  return text.replaceAll(RegExp(r'[^0-9]'), '').length;
}
Hence there should be 0 in the output. I have no clue what happens.

t-m-z:
The easy way is I remove this line  😉

 

Von: Mike-3 ***@***.***> 
Gesendet: Tuesday, 11 May 2021 18:48
Betreff: Re: [S-Man42/GCWizard] Tmz 1.6.0 divisor of a number (#515)

 

I think I see the problem
GCWCrosstotalOutput(text: _currentInputN.toString()
_currentInputN is a int. And now you have 0 in text.

Mike-3:
Count Length problem unsolved

t-m-z:
> Count Length problem unsolved

sorry - forget the PR - however, there is still the problem with the empty integer spinner, who returns a 0, not a null.

S-Man42:
What about this. Is there still a problem?

t-m-z:
what is your opinion?

Currently an empty result from the Integer Spinner is treated as Zero, not null.
I could live with that.

Sent from MailDroid

-----Original Message-----

S-Man42:
Yes, this is intended and designed like that. 

Example:

You have an integer spinner for a value you simply want to add. Maybe you have the value "123" and want to change it to "234". Currently your output show 100 + 123 = 223. Then you have to delete the 3 (output shows 100 + 12 = 112), then the 2 (output shows 100 + 1 = 101) and then? Should it crash because of the empty input? In most cases the tools would catch the null and treat it as 0 in that case. Ok, you can give out an error message. But I am no friend of unnecessary error messages because the make the user feel uncomfortable.

Moreover it makes some tools easier to use (especially just in the role of a tester). You don't need to fill out all fields, because internally all fields are still filled out with 0. (e.g. the many inputs of a DMM or DMS coord format, where it becomes very useful in tests that you don't need to fill out it explicitely).

Last thing: If you made it NULL, what should happen if you click +1/-1 buttons in that case? At the moment it assumes, that there is value 0. So the buttons have no problem at all. What if it was NULL? Do the buttons crash/return error? Or treat they the NULL as zero? I decided second because it's more useful, I guess ;)

So the behaviour of the spinners or number input fields shouldnt be changed anymore. Does your tool has a problem with that?

t-m-z:
> Yes, this is intended and designed like that.
> 
> Example:
> 
> You have an integer spinner for a value you simply want to add. Maybe you have the value "123" and want to change it to "234". Currently your output show 100 + 123 = 223. Then you have to delete the 3 (output shows 100 + 12 = 112), then the 2 (output shows 100 + 1 = 101) and then? Should it crash because of the empty input? In most cases the tools would catch the null and treat it as 0 in that case. Ok, you can give out an error message. But I am no friend of unnecessary error messages because the make the user feel uncomfortable.
> 
> Moreover it makes some tools easier to use (especially just in the role of a tester). You don't need to fill out all fields, because internally all fields are still filled out with 0. (e.g. the many inputs of a DMM or DMS coord format, where it becomes very useful in tests that you don't need to fill out it explicitely).
> 
> Last thing: If you made it NULL, what should happen if you click +1/-1 buttons in that case? At the moment it assumes, that there is value 0. So the buttons have no problem at all. What if it was NULL? Do the buttons crash/return error? Or treat they the NULL as zero? I decided second because it's more useful, I guess ;)
> 
> So the behaviour of the spinners or number input fields shouldnt be changed anymore. Does your tool has a problem with that?

the tool has no problwm - it was Mike who saw a problem with an empty spinner

lib/widgets/common/gcw_crosstotal_output.dart
```diff
@@ -24,10 +25,18 @@ class _GCWCrosstotalOutputState extends State<GCWCrosstotalOutput> {
     var text = widget.text;
     List<int> values = List.from(widget.values);
 
-    var crosstotalValues = [
+    var crosstotalValues;
+    if (widget.divisor)
```
S-Man42, line 29: Oh just saw this.

I don't think, it's a good idea to couple a very common widget with a very specific tool. I see the necessarity here but the name of the member field is not good. Maybe something like `suppressCharacterCounts` and then simply:

`widget.suppressSums ? null : ...`
to
`widget.suppressSums || widget.suppressCharacterCounts ? null : ...`

for the first two entries instead of the `if `construct

t-m-z: I see your point - but it is not only the suppression of lines but also a renaming of the output.
I will check

t-m-z: rename is done, but I kept the if construct

PR #516, opened 2021-04-30
Tmz 1.6.0 complex numbers

t-m-z:
converts between cartesian and polar coordinates


assets/i18n/en.json
```diff
@@ -974,7 +974,17 @@
   "common_unit_volume_smi_name": "US Smidgen",
   "common_unit_volume_pn_name": "US Pinch",
   "common_unit_volume_ds_name": "US Dash",
-  "coords_selection_title": "Coordinate Calculation",
```
Mike-3, line 977: not remove coords_selection_title

Mike-3 (commented):
character input posibile. Please change, if possible.
Possibly, for example, 1/3 to be accepted as a valid input (use math_expressions)

Mike-3 (approved):

S-Man42:
Same issue as on quadratic: Please add some space between the inputs :)

Another thing I saw: Why can I enter any character? I guess there should be an input mask for only [\-0-9,.]. Don't you think?

t-m-z:
> Same issue as on quadratic: Please add some space between the inputs :)

Layout issue is solved

> Another thing I saw: Why can I enter any character? I guess there should be an input mask for only [-0-9,.]. Don't you think?

Hm - I used your DoubleTextInputFormatter - maybe there is a bug in the regex?

t-m-z:
PR done

S-Man42:
I'll check the input issue...

PR #520, opened 2021-05-02
Tmz 1.6.0 navajo

t-m-z:
encoding and decoding using the Navajo Code Talkers' Dictionary REVISED 15 JUNE 1945 (DECLASSIFIED UNDER DEPARTMENT OF DEFENSE DIRECTIVE 5200.9)
from 
https://www.history.navy.mil/research/library/online-reading-room/title-list-alphabetically/n/navajo-code-talker-dictionary.html
https://www.ancestrycdn.com/aa-k12/1112/assets/Navajo-Code-Talkers-dictionary.pdf

tests checked with 
https://kryptografie.de/kryptografie/chiffre/navajo.htm
https://www.dcode.fr/navajo-code

hint
1. the alphabet-table at kryptografie.de has errors - misspelling
2. the whole dictionary is not implemented - only the alphabet
3. kryptografie uses a short version of the alphabet
4. dcode uses the alphabet table from the dictionary and encodes letters randomly


lib/logic/tools/crypto_and_encodings/navajo.dart
```diff
@@ -0,0 +1,823 @@
...
+Map NAVAJO_DECODE_DICTIONARY = switchMapKeyValue(NAVAJO_ENCODE_DICTIONARY);
+
+Map NAVAJO_DECODE_ALPHABET = switchMapKeyValue(NAVAJO_ENCODE_ALPHABET);
+
+String shrinkText(String input){
+  return input
+      .replaceAll('COMMANDING GEN.', 'COMMANDINGGEN')
+      .replaceAll('MAJOR GEN.', 'MAJORGEN')
+      .replaceAll('BRIGADIER GEN.', 'BRIGADIERGEN')
+      .replaceAll('LT. COLONEL', 'LTCOLONEL')
```
S-Man42, line 665: IIN line 51 you write LT.COLONEL with a dot... Problem?

t-m-z: checked and changed

S-Man42:
Several issues here:

1. It's not clear to what "only alphabet" does. Can you make an example please? "Only Alphabet" only returns the input....
grafik

2. Click on second Switch directly after opening the tool: Both switches move, not only the second. They only become independent when you switch the first one first
3. When typing an input, and then switching the source removes the input... Seems not very comfortable...
4. Some strange misbehaviour occurs when play long enough with the switches. In this example I guess it's the output from encryption not from decryption:
grafik

I guess your state handling is not proper.

S-Man42:
Ah, I believe I found the problem: Your second switch sets _currentMode not _currentSource

t-m-z:
1. Behaviour of 2nd switch fixed
2. change "Source" to "Using" - this should make it better to understand
3. Switching between encrypt and decrypt does also switch to a new controller. Hence the input seems to be deleted. But this behaviour is found in several tools

PR done

lib/logic/tools/crypto_and_encodings/navajo.dart
```diff
@@ -0,0 +1,823 @@
...
+  NAVAJO_DECODE_ALPHABET['SHI-DA'] = 'U';
+  NAVAJO_DECODE_ALPHABET['BSEH-DO-GLIZ'] = 'Z';
+
+  List<String> result = new List<String>();
+  if (cipherText == null || cipherText == '') return '';
+
+  cipherText.toUpperCase().split(' ').forEach((element) {
+    if (NAVAJO_DECODE_ALPHABET[element] == null)
+      if (useOnlyAlphabet)
+        result.add('?');
```
S-Man42, line 785: Here two options:

1. Normally we ignore input which do not deliver a result. For example "AAA 111 222" simply ignores the "AAA" block in Vanity and does not return "unknown". The idea behind is to enter any text and return only the relevant information for direct copy without the need to remove some "?" chars afterwards manually. Think about a listing which is: "And navajo indian said: A-WOH SHI-DA, but I didn't understand him". In that case you could input the entire text and should only return "TU". I know, there are tools that behave differently, like the Alphabet Values which signalize that "Ä" is not part of the chosen Standard Alphabet or something. But in your tool I would tend to the first way
2. However, if you decide for the second way, please use the common constant "UNKNOWN_ELEMENT" instead.

lib/logic/tools/crypto_and_encodings/navajo.dart
```diff
@@ -0,0 +1,823 @@
+// https://www.history.navy.mil/research/library/online-reading-room/title-list-alphabetically/n/navajo-code-talker-dictionary.html
+// https://www.ancestrycdn.com/aa-k12/1112/assets/Navajo-Code-Talkers-dictionary.pdf
+// https://kryptografie.de/kryptografie/chiffre/navajo.htm
```
S-Man42, line 3: please DO NOT use kryptografie.de as source!

S-Man42:
Input Encrypt:
"sdf sdf" -> "DIBEH BE MA-E  DIBEH BE MA-E"

Taking this output as input Decrypt:
"DIBEH BE MA-E  DIBEH BE MA-E" -> "SDF?SDF". 

The space becomes a "?".

The encrypted output now has two spaces. I guess you make something like "inputtext".split(' '). In that case you create and empty object (between the two spaces). Better sanitize your input. Two options:
1.` input.replaceAll('\s+', ' ')` // reduces all whitespaces, no matter how long to one single space before processing
2. `input.split(' ').where((element) => element != null && !element.isEmpty())...` // removes the empty elements from split result

A third (maybe additional) option can be to ignore the space in encrypt output. Is it really necessary to return a double space? ;)

lib/logic/tools/crypto_and_encodings/navajo.dart
t-m-z: done

lib/widgets/searchstrings/searchstrings_common.dart
```diff
@@ -205,6 +205,7 @@ const SEARCHSTRING_COMMON_MAYANUMBERS = 'mayas majas vigesimalsystem 20';
 const SEARCHSTRING_COMMON_MEXICANARMYCIPHERWHEEL = '  ';
 const SEARCHSTRING_COMMON_MORSE = 'samuel morse ';
 const SEARCHSTRING_COMMON_MULTIDECODER = '';
+const SEARCHSTRING_COMMON_NAVAJO = 'navajo';
```
S-Man42, line 208: i'd add "najaho" as well. Saw that sometimes (for example my old climbing gear is called "Petzl Navaho" - they call their gear after indian tribes)

t-m-z: check - done

t-m-z:
> DIBEH BE MA-E DIBEH BE MA-E

Good point - the idea behind: how to deal with words encoded letterwise? The double whitespace should symbolize, there starts the next word.
One option could be to simply the process and ignore these spaces.
One option could be to make it more sophisticated.

I will think about.

lib/logic/tools/crypto_and_encodings/navajo.dart
t-m-z: done

S-Man42:
KISS: Keep it simple and stupid :)

S-Man42:
Decision here? Can this be merged or is it WIP?

t-m-z:
can be merged

Sent from MailDroid

-----Original Message-----

PR #523, opened 2021-05-03
Hex2file

Mike-3:


Mike-3:
short docu ftp://Exchange/HexString2File.docx

lib/widgets/tools/science_and_technology/hexstring2file.dart
```diff
@@ -0,0 +1,179 @@
...
+
+      if (_outData == null)
+        return null;
+
+      var mimeType = getMimeType(getFileType(_outData));
+
+      switch (mimeType) {
+        case MIMETYPE.IMAGE:
+          try {
+            return GCWSymbolContainer(
```
S-Man42, line 106: A Symbolcontainer shouldn't be used here... It is meant to be used only for symbols of the symbol tables... So if once the symbol layout may change, this affects here (e.g. adding a background color or a border or something.)

PR #527, opened 2021-05-03
Tmz 1.6.0 fix beaufort table

t-m-z:
adjust symbols according to current definition of WMO


t-m-z:
something seems wrong

PR #534, opened 2021-05-06
Registry perf2

ludoo0d0a:
push new clean branch for PR (based on 1.6.0 properly)


S-Man42:
That looks better, but why is the Exif code in here. Was this intended?

Please close the PRs that are not needed anymore :)

lib/logic/tools/coords/data/coordinates.dart
```diff
@@ -98,6 +98,18 @@ String _getSignString(int sign, bool isLatitude) {
   return _sign;
 }
 
+int getSignFromString(String text, bool isLatitude) {
+  int _sign = 0;
+
+  if (isLatitude) {
+    _sign = (text == 'N') ? 1 : -1;
+  } else {
+    _sign = (text == 'S') ? 1 : -1;
```
S-Man42, line 107: "S" needs to be "E" here.

lib/logic/tools/coords/data/coordinates.dart
```diff
@@ -98,6 +98,18 @@ String _getSignString(int sign, bool isLatitude) {
   return _sign;
 }
 
+int getSignFromString(String text, bool isLatitude) {
```
S-Man42, line 101: `getCoordinateSignFromString` would be a better name. Because this is a public method, it is necessary to create a name which exactly points to the context. Otherwise it could happen that someone uses it for the wrong task ("sign string" could be used in several contextes)

lib/logic/tools/crypto_and_encodings/exif_reader.dart
```diff
@@ -0,0 +1,150 @@
...
+    LatLng _point = decToLatLon(DEC(_lat, _lng));
+
+    print("_point = ${_point}");
+    return _point;
+  }
+
+  return null;
+}
+
+double getCoordDec(IfdTag tag, String latlngRef, bool isLatitude) {
```
S-Man42, line 62: You always created public methods. Please make them private if they are not used anywhere else to keep the number of publicly usable methods as small as possible.

lib/logic/tools/crypto_and_encodings/exif_reader.dart
```diff
@@ -0,0 +1,150 @@
...
+}
+
+int getExifRotation(Map<String, IfdTag> tags) {
+  IfdTag orientation = tags["Image Orientation"];
+  int orientationValue = orientation.values[0];
+  int rotationCorrection = 0;
+  // in degress
+  print("orientation: ${orientation.printable}/${orientation.values[0]}");
+  switch (orientationValue) {
+    case 6:
```
S-Man42, line 138: what's with the other values?

lib/widgets/registry.dart
```diff
@@ -708,21 +709,33 @@ class Registry {
...
+            SEARCHSTRING_FR_FORMULASOLVER,
           ]),
+
       GCWTool(tool: FormulaSolverFormulaGroups(), i18nPrefix: 'formulasolver', missingHelpLocales: [], searchStrings: [
         SEARCHSTRING_COMMON_FORMULASOLVER,
         SEARCHSTRING_DE_FORMULASOLVER,
         SEARCHSTRING_EN_FORMULASOLVER,
-        SEARCHSTRING_FR_FORMULASOLVER
-      ]),
+        SEARCHSTRING_FR_FORMULASOLVER,
```
S-Man42, line 731: why did you add the commas?

lib/widgets/searchstrings/searchstrings_common.dart
```diff
@@ -129,6 +129,7 @@ const SEARCHSTRING_COMMON_ESOTERIC_DEADFISH = 'deadfish idso xkcd ';
 const SEARCHSTRING_COMMON_ESOTERIC_MALBOLGE = 'malbolge olmstead';
 const SEARCHSTRING_COMMON_ESOTERIC_OOK = 'ook terry pratchett  ';
 const SEARCHSTRING_COMMON_ESOTERIC_WHITESPACELANGUAGE = 'whitespace';
+const SEARCHSTRING_COMMON_EXIF = 'exif';
```
S-Man42, line 132: other useful strings are
GERMAN: metadaten bilder
ENGLISH: metadata images pictures
FRENCH: ...

lib/widgets/registry.dart
```diff
@@ -708,21 +709,33 @@ class Registry {
...
-            SEARCHSTRING_FR_ESOTERICPROGRAMMINGLANGUAGE
+            SEARCHSTRING_FR_ESOTERICPROGRAMMINGLANGUAGE,
+          ]),
+      GCWTool(
+          tool: ExifReader(),
+          i18nPrefix: 'exif',
+          category: ToolCategory.SCIENCE_AND_TECHNOLOGY,
+          missingHelpLocales: [],
+          searchStrings: [SEARCHSTRING_COMMON_EXIF,
+            SEARCHSTRING_DE_FORMULASOLVER,
```
S-Man42, line 722: FORMULASOLVER searchstrings are wrong here :)

S-Man42:
Please add tests for the logic. Especially the GPS Section would be very interesting since there are many edge cases to catch (-+90 als limits, so what happens if deg == 90 and minutes == 1? --> there's a normalize function in the coordinate sections... moreover, I'd tend to convert the coords into LatLon using the appropriate function, which already do such checks...; what havens if minutes >= 60, ...)

ludoo0d0a:
I made really bad work .. i hate git branching
i need to clean up
exif has nothing to do here. bad cherry pick

lib/logic/tools/crypto_and_encodings/exif_reader.dart
ludoo0d0a: no exif feature in this branch

lib/widgets/registry.dart
ludoo0d0a: no exif feature in this branch

lib/widgets/searchstrings/searchstrings_common.dart
ludoo0d0a: no exif feature in this branch

lib/logic/tools/coords/data/coordinates.dart
ludoo0d0a: no exif feature in this branch

lib/logic/tools/coords/data/coordinates.dart
ludoo0d0a: no exif feature in this branch

lib/widgets/registry.dart
ludoo0d0a: no exif feature in this branch

lib/logic/tools/crypto_and_encodings/exif_reader.dart
ludoo0d0a: no exif feature in this branch

PR #538, opened 2021-05-09
Registry perf3

ludoo0d0a:
finally my PR for search engine optimization

proper branch from last 1.6.0, and cherry pick + cleanup

(sorry again for bad quality)

### DIFF
--- a/lib/utils/common_utils.dart
+++ b/lib/utils/common_utils.dart
@@ -88,6 +88,12 @@ String normalizeUmlauts(String input) {
   }).join();
 }
 
+List<String> splitWords(String text){
+  // TODO Move regex static/const
+  RegExp reSplit = RegExp(r'[\s,]');
+  return text.split(reSplit);
+}
+
 String removeAccents(String text) {
   String out = normalizeUmlauts(text);
   return removeDiacritics(out);
--- a/lib/utils/common_utils.dart
+++ b/lib/utils/common_utils.dart
@@ -88,19 +88,15 @@ String normalizeUmlauts(String input) {
   }).join();
 }
 
-List<String> splitWords(String text){
-  // TODO Move regex static/const
-  RegExp reSplit = RegExp(r'[\s,]');
-  return text.split(reSplit);
-}
-
 String removeAccents(String text) {
   String out = normalizeUmlauts(text);
   return removeDiacritics(out);
 }
 
+final RegExp reNonLetters = RegExp(r'[^A-Za-z]');
+
 String removeNonLetters(String text) {
-  return text.replaceAll(RegExp(r'[^A-Za-z]'), '');
+  return text.replaceAll(reNonLetters, '');
 }
 
 String insertCharacter(String text, int index, String character) {
--- a/lib/widgets/common/gcw_tool.dart
+++ b/lib/widgets/common/gcw_tool.dart
@@ -87,6 +87,7 @@ class GCWTool extends StatefulWidget {
   final autoScroll;
   final iconPath;
   final List<String> searchStrings;
+  List<String> _indexedStrings;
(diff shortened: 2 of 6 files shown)


lib/widgets/registry.dart
```diff
@@ -6263,5 +6267,21 @@ class Registry {
...
+    buildIndexedStrings();
+
+  }
+
+  // Build indexed strings : concatenated lower case no accent
+  static void buildIndexedStrings() {
+    indexedTools = toolList.where((tool) {
+      var _indexedStrings = removeAccents(tool.searchStrings.join(' ').toLowerCase());
+      if (_indexedStrings == null || _indexedStrings.length == 0) return false;
+      tool.indexedStrings = _indexedStrings;
```
S-Man42, line 6283: `tool.indexedStrings` is not defined.... You added private `_indexedStrings` to `GCWTool` instead. I guess this needs to be public instead?

lib/widgets/common/gcw_tool.dart
```diff
@@ -124,6 +125,12 @@ class GCWTool extends StatefulWidget {
     }
   }
 
+  String get indexedStrings => _indexedStrings;
```
S-Man42, line 128: Why not simply make _indexedStrings (and remove explicit getter and setter) to indexedStrings? AFAIK this results in the same behaviour. Am I wrong?

S-Man42:
Some minor things (see comments). Looks good so far nonetheless: Great work! :)

S-Man42:
One enhancement comes to my mind: Remove duplicate values: If, for example, there's a word used in two searchstring files (maybe COMMON and ENGLISH or theres the same word in two languages), you don't need to add both... So the best would be to collect all string in a Set (=> removes duplicates) and convert this to a List afterwards

lib/widgets/main_view.dart
```diff
@@ -462,27 +464,21 @@ class _MainViewState extends State<MainView> {
...
   List<GCWTool> _getSearchedList() {
-    var list = Registry.toolList;
-    String searchstring = '';
-
-    list = list.where((tool) {
-      searchstring = tool.searchStrings.join(' ').toLowerCase();
-      if (searchstring == null || searchstring.length == 0) return false;
-
-      var found = true;
+    Stopwatch stopwatch = Stopwatch()..start();
```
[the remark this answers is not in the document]
ludoo0d0a: I waited the last moment. I wasn't sure.
ok I will remove it then

lib/widgets/common/gcw_tool.dart
ludoo0d0a: getter only, no setter 
but yes, more or less the same result

ludoo0d0a: mmh, I was wrong. I use setter... Moved to public field (i just discovered that pair getter+setter is marked as unnecessary wrapper warning in intellij )

ludoo0d0a:
> One enhancement comes to my mind: Remove duplicate values: If, for example, there's a word used in two searchstring files (maybe COMMON and ENGLISH or theres the same word in two languages), you don't need to add both... So the best would be to collect all string in a Set (=> removes duplicates) and convert this to a List afterwards

removeDupilcates added

lib/utils/common_utils.dart
```diff
@@ -88,6 +88,12 @@ String normalizeUmlauts(String input) {
   }).join();
 }
 
+List<String> splitWords(String text){
```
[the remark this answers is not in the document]
ludoo0d0a: splitWords moved to Registry.splitWordsOnSpace

lib/widgets/main_view.dart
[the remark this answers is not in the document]
ludoo0d0a: removed

PR #546, opened 2021-05-15
columnedMultiLineOutput - copy all values option

Mike-3:


lib/widgets/utils/common_widget_utils.dart
```diff
@@ -37,7 +37,7 @@ defaultFontSize() {
 }
 
 List<Widget> columnedMultiLineOutput(BuildContext context, List<List<dynamic>> data,
-    {List<int> flexValues = const [], int copyColumn, hasHeader: false, List<Function> tappables}) {
+    {List<int> flexValues = const [], int copyColumn, bool hasHeader: false, bool copyValueColumn: false, List<Function> tappables}) {
```
S-Man42, line 40: I don't like the name :D Maybe better: copyAll

lib/widgets/utils/common_widget_utils.dart
```diff
@@ -62,6 +62,12 @@ List<Widget> columnedMultiLineOutput(BuildContext context, List<List<dynamic>> d
 
     if (copyColumn == null) copyColumn = rowData.length - 1;
     var copyText = rowData[copyColumn].toString();
+    if ((isFirst && hasHeader) & copyValueColumn) {
```
S-Man42, line 65: Is te & operator correct here (and below). Shouldn't it be: &&

Mike-3: Why not ?
(isFirst && hasHeader) is original, 
I can change this to (isFirst & hasHeader & copyValueColumn)

lib/widgets/utils/common_widget_utils.dart
Mike-3: I also had it first. But only the Value column is copied. If everything has to be copied, it doesn't fit. 
copyValues ?

lib/widgets/utils/common_widget_utils.dart
S-Man42: To be honest, I am not familiar with the bitwise AND. Does it really do the same like &&? 

I'd write 

`isFirst && hasHeader && copyValueColumn`

lib/widgets/utils/common_widget_utils.dart
S-Man42: copyValues is not better... copyAllCopyValues? :D Hm, I believe "copyAll" is not correct, sure - but it's close enough...

PR #561, opened 2021-05-21
close #555 Tmz shadoks graphic encoding/decoding

### LINKED ISSUE #555: [New feature] Shadoks numbers
Add a new tool to provide shadok numbers conversion

https://www.dcode.fr/numeration-shadok

(idea from daniel.b)

t-m-z:


S-Man42:
I'd add a symbol table shadoks as well: if someone only sees the symbols without knowing that this is shadoks, he would search in the symbol tables first. What do you think? As we have for Cistercian, Babylonian and Maya as well.

Then you can create a Shadoks Selection with both, the symbol table and the decoder tool.
(Btw. can you add the Maya Calender to the Maya Selection as well? Then it can be found under Maya AND under Date Functions)

S-Man42:
grafik
See the difference between the large image and the small output image.

Furthermore I cannot click the circle at the right side...

lib/widgets/tools/crypto_and_encodings/shadoks_numbers_segment_display.dart
```diff
@@ -0,0 +1,99 @@
...
+              paint.color =
+                  currentSegments['a'] ? SEGMENTS_COLOR_ON : SEGMENTS_COLOR_OFF;
+              paint.style = PaintingStyle.stroke;
+              paint.strokeWidth = 5.0;
+              canvas.drawCircle(
+                  Offset(size.width / SEGMENTS_RELATIVE_DISPLAY_WIDTH * 48,
+                      size.height / SEGMENTS_RELATIVE_DISPLAY_HEIGHT * 48),
+                  size.height / SEGMENTS_RELATIVE_DISPLAY_HEIGHT * 43,
+                  paint, onTapDown: (tapDetail) {
+                setSegmentState('a', !currentSegments['a']);
```
S-Man42, line 39: To get the behaviour: Circle of, if b,c,d is clicked you simply have to add this here:
```
setSegmentState('a', !currentSegments['a']);
setSegmentState('b', false);
setSegmentState('c', false);
setSegmentState('d', false);
```

lib/widgets/tools/crypto_and_encodings/shadoks_numbers_segment_display.dart
```diff
@@ -0,0 +1,99 @@
...
+              pathB.lineTo(size.width / SEGMENTS_RELATIVE_DISPLAY_WIDTH * 74,
+                  size.height / SEGMENTS_RELATIVE_DISPLAY_HEIGHT * 15);
+              pathB.lineTo(size.width / SEGMENTS_RELATIVE_DISPLAY_WIDTH * 74,
+                  size.height / SEGMENTS_RELATIVE_DISPLAY_HEIGHT * 81);
+              pathB.lineTo(size.width / SEGMENTS_RELATIVE_DISPLAY_WIDTH * 64,
+                  size.height / SEGMENTS_RELATIVE_DISPLAY_HEIGHT * 85);
+              pathB.close();
+
+              canvas.drawPath(pathB, paint, onTapDown: (tapDetail) {
+                setSegmentState('b', !currentSegments['b']);
```
S-Man42, line 58:
```
setSegmentState('b', !currentSegments['b']);
setSegmentState('a', false);
```
(as well as at c and d, too, of course)

t-m-z:
> I'd add a symbol table shadoks as well: if someone only sees the symbols without knowing that this is shadoks, he would search in the symbol tables first. What do you think? As we have for Cistercian, Babylonian and Maya as well.
> 
> Then you can create a Shadoks Selection with both, the symbol table and the decoder tool.
> (Btw. can you add the Maya Calender to the Maya Selection as well? Then it can be found under Maya AND under Date Functions)


both added to the to do list

t-m-z:
> grafik
> See the difference between the large image and the small output image.
> 
> Furthermore I cannot click the circle at the right side...

The difference between the two images seem to be a scaling-problem.
Regarding the circle on the right side, I can reproduce this. It seems the whole area does not recognize tapping.

S-Man42:
No, it's definitely no scaling issue. It's simply a different handling of the input and output! ;)

S-Man42:
Encryption does not work:
shadoks

S-Man42:
Decryption Gif:
shadoks

You can see a complete misbehaviour, however. If it was a scaling problem, some segments wouldn't be shown anytime. But you can see, that in some cases they are shown.

S-Man42:
Moreover I would make this tool as flexible as the other segment tools:
1 = any of the three is checked
2 = any two of the three are checked.

t-m-z:
> Moreover I would make this tool as flexible as the other segment tools:
> 1 = any of the three is checked
> 2 = any two of the three are checked.

I thought of that but this would not stick to the standard although it is easier to implement. So I decided not to do.

PR #568, opened 2021-05-23
add missing classes

t-m-z:
sorry for that lack of knowledge


lib/widgets/main_view.dart
```diff
@@ -114,6 +115,7 @@ import 'package:gc_wizard/widgets/tools/crypto_and_encodings/rotation/rot18.dart
 import 'package:gc_wizard/widgets/tools/crypto_and_encodings/rotation/rot47.dart';
 import 'package:gc_wizard/widgets/tools/crypto_and_encodings/rotation/rot5.dart';
 import 'package:gc_wizard/widgets/tools/crypto_and_encodings/rotation/rotation_general.dart';
+import 'package:gc_wizard/widgets/tools/crypto_and_encodings/shadoks_numbers.dart';
```
S-Man42, line 118: I'd leave that because I already added the Shadoks Selection. This is enough I think.

lib/widgets/main_view.dart
```diff
@@ -99,6 +99,7 @@ import 'package:gc_wizard/widgets/tools/crypto_and_encodings/language_games/duck
 import 'package:gc_wizard/widgets/tools/crypto_and_encodings/language_games/pig_latin.dart';
 import 'package:gc_wizard/widgets/tools/crypto_and_encodings/language_games/robber_language.dart';
 import 'package:gc_wizard/widgets/tools/crypto_and_encodings/language_games/spoon_language.dart';
+import 'package:gc_wizard/widgets/tools/crypto_and_encodings/maya_numbers.dart';
```
S-Man42, line 102: Should we put the Maya Calendar under "Maya Numbers" as well? In that case we just need the maya_numbers_selection

lib/widgets/main_view.dart
```diff
@@ -172,6 +175,8 @@ import 'package:gc_wizard/widgets/tools/science_and_technology/segment_display/f
 import 'package:gc_wizard/widgets/tools/science_and_technology/segment_display/seven_segments.dart';
 import 'package:gc_wizard/widgets/tools/science_and_technology/segment_display/sixteen_segments.dart';
 import 'package:gc_wizard/widgets/tools/science_and_technology/unit_converter.dart';
+import 'package:gc_wizard/widgets/tools/science_and_technology/vanity_words_list.dart';
```
S-Man42, line 178: Hm... Not quite sure here as well...

I know, it may be confusing. 

Currently the logic is: 
- Complete independent tools are listed in the main_view
- If there's a tool with own subsections, only the selections are added... 

It's just a type of... "stomach" decision (you know, "Bauchentscheidung"). It should be a more or less complete overview but no-one is happy with every single subsection as entry. So we need to make a decision, what should be used here and what not... Please decide on your own :)

S-Man42:
What about your mathematical functions? These should be listed without any doubt :)

lib/widgets/main_view.dart
t-m-z: ok - I understand

lib/widgets/main_view.dart
t-m-z: I would say the Maya(Numbers/Calendar)Selecten is sufficient

lib/widgets/main_view.dart
t-m-z: check - so I will keep only the selection

t-m-z:
> What about your mathematical functions? These should be listed without any doubt :)

QuadraticEquation is alread listed
ComplexNamubers is alread listed
SilverRatioSelection is already listed

t-m-z:
Done

Issue #475, opened 2021-04-12
Steganography support


ludoo0d0a:
steganography can be support , if we can read image in exif module

to start : 
https://pub.dev/packages/flutter_stegify
https://pub.dev/packages/flutter_steganography

S-Man42:
Looks great. If you like, please have a look at #411 (my personal favorite crypto, I'd love to see it in the GCW) or #412 o as well #478

ludoo0d0a:
done in exif_reader PR
https://github.com/ludoo0d0a/GCWizard/commit/c510d04694a278b14416d7f4a3e9ae40a2c0302f

and added filetype (BOM) as well

S-Man42:
Can this be closed? When I remember correctly, you wanted to enhance the feature

PR #569, opened 2021-05-24
Steganography

ludoo0d0a:
Steganography in image ( #475 )

- save in download folder (to be discussed)
- textfields uses labelText, not hintText
- plugin lib/plugins/flutter_steganography pasted to manage dependency, and image outputs (may be evolved if onwer allow evolutions)


S-Man42:
conflicts here as well

ludoo0d0a:
> conflicts here as well

fixed

pubspec.yaml
```diff
@@ -10,7 +10,9 @@ dependencies:
   auto_size_text: ^2.1.0
   base32: ^2.0.0-nullsafety.1
   cached_network_image: ^3.0.0
+  ext_storage: ^1.0.3
```
S-Man42, line 13: we now have different file pickers. Am I right? Is this intended?

S-Man42:
You know, I feel I bit uncomfortable with the simple name "Steganography". In fact, the used algorithm is the "Least Significant Bit Steganography" (acc. to the README of the flutter_steganography plugin), which, indeed, is a very common algorithm. So please rename it to something more fitting, e.g.

* Steganography (Least Significant Bit LSB)
* LSB Steganography
* whatever...

Add "leastsignificantbit lsb" to the searchstrings

Please, rename the i18n prefix to "lsb_steganography"

S-Man42:
Did you already cross-checked if your tool works when the input is encrypted with another external tool which does LSB? The goal must be that a geocacher finds a certain image in a listing - and we can assume that this image was created by another tool than the GCW - and can successful decode the image with our tool :) That's why it is important to know exactly the used algorithm, because there are many different stego algorithms out there.

https://incoherency.co.uk/image-steganography/
https://medium.com/ctf-writeups/stegonline-a-new-steganography-tool-b4eddb8f8f57
https://stylesuxx.github.io/steganography/
https://github.com/RobinDavid/LSB-Steganography

lib/widgets/searchstrings/searchstrings_common.dart
```diff
@@ -190,6 +190,7 @@ const SEARCHSTRING_COMMON_ICECODES = '1980 wmo sigrid';
 const SEARCHSTRING_COMMON_IMAGE_COLORCORRECTIONS = '';
 const SEARCHSTRING_COMMON_IMAGES = '';
 const SEARCHSTRING_COMMON_IMAGESANDFILESSELECTION = '';
+const SEARCHSTRING_COMMON_STEGANO = 'stegano steganography';
```
S-Man42, line 193: Just for your interest: "steganography" contains "stegano", so it is not necessary to write both keywords. The search looks for word parts and so will find "stegano" as well.

Moreover: "steganography" is an English term, so I'd put it into SEARCHSTING_EN instead of _COMMON

lib/widgets/utils/file_utils.dart
```diff
@@ -1,9 +1,9 @@
 import 'dart:io';
 import 'dart:typed_data';
```
S-Man42, line 2: please have a look at this file. Currently (mikes current version) does not use ext_storage anymore, you embed this here. Additionally the method "mainDirectory" is commented (not used) in mikes version. Please ensure, that you both use the same file. We MUST avoid that you both play ping pong with the file...

pubspec.yaml
S-Man42: I guess, we decided for flutter_writeable, didn't we?

pubspec.yaml
```diff
@@ -10,7 +10,9 @@ dependencies:
...
   diacritic: ^0.1.1
+  encrypt: ^5.0.0
```
S-Man42, line 15: As I see: "encrypt" is a framework on top of "Pointy Castle". Currently we are using pointycastle explicitely for the hash function. Can we remove this when we use encrypt now? Want to avoid using the lib twice...

PR #616, opened 2021-06-04
ios gpx filter

Mike-3:


S-Man42:
KML does not work, too...

However, it's interesting. While on my device these files are disabled, on Andys it is working correctly... (without your fix here...). So I am not sure if this is really necessary - are you?

Mike-3:
In this case, is this Merge mit nessecery.

PR #639, opened 2021-06-24
Visual Cryptography

Mike-3:
bugfix imageview (filename not used)


S-Man42:
Works great for me! Very well done.

Now we need to talk about the UI :D

1. Save: In my example the files were saved without .png suffix, so the loader couldnt find it.
2. Please add some space between the two decryption images: 
grafik
3. Please display the file name of the loaded files, so one can see, which one is already opened
4. What does "Auto XY" and "Auto X" exactly do? First one is horrible slow (stucking at 0%), second one delivered nothing after slow calculation...
5. The idea with the extra output button is nice, but indeed it was not clear to me, what that meant. Maybe it would be better to add the finalized output automatically under the overlayed image? Or simply add a button: "Show clear image" or something?

S-Man42:
Ich habs gerade tatsächlich gebraucht (sollte nen Rätsel für nen Freund erstellen).

Dabei ist mir aufgefallen, dass die Bilder sehr groß werden. Hatte nen einfachen QR Code, der wurde dann in 800x800 px zerlegt. Das ist schon schwierig, den auszudrucken (darf man ja nicht skalieren wegen Unschärfen). Cool wäre also beim Create noch ne Option: Ausgabegröße. Ggf. muss das Inputbild dann eben nochmal skaliert werden. Aber bei nem QR Code mit ca. 30x30 pixeln sollte ja ein Bild mit 60x60 statt 800x800 Minimum machbar sein, oder? :)

Mike-3:
1. Save: In my example the files were saved without .png suffix, so the loader couldnt find it.
- actual file_picker used ? input box from file picker limeted  of 30 characters. 
The generated filename is limited to 30 characters in the new version. 

2. However, I am not even sure, if you need these preview images, which always show some gray boxes... maybe only the filename?
 - ok, only file name

4.    What does "Auto XY" and "Auto X" exactly do? First one is horrible slow (stucking at 0%), second one delivered nothing after slow calculation...
- tries to automatically determine the offset in XY or only X in the case of shifted images. The determined value is entered in the offsets. If it was 0 and 0 was determined, e.g. no change can be seen.
The investigation really takes a long time. 

5. ok, make an extra button. Just wanted to save translations.

Mike-3:
I cannot understand the problem with the image size. See the appendix for an example.
At most a large offset has been entered. 
QuickResponse
qr1
qr2

Mike-3:
1. Save: In my example the files were saved without .png suffix, so the loader couldnt find it.
- ok, i see the problem. My last changes have been overwritten.

S-Man42:
Concerning "Scale":
1. Why percent instead of real expected output? I am not quite sure, maybe could work also. I was just wondering that you decided % instead of width-px as we chose in the call... Well, I guess, it's ok. Maybe you can add an information line with the expected width/height for chosen scale?
2. Default is scale = 0%. Shouldn't it be: 100% (original size). 50% would be half the size and 200% doubled (per side of course). This is how all image tools work.
3. Changing the scale value affects the offsetY value in decrypt. I guess you use the same variable for both inputs. Please correct.

Furthermore:
I created an output with 50% and one with 100%. Then for tests I used one image from 50 and one for 100. In my expectation one quarter of the resulting image should be darker (the one where the half size overlaps the full size) and the rest should look quite normal. But I got this:
 
grafik

Its the other way round. The rest isn't 50/50 black/white but completely black. How can this be explained. Is this a bug or a feature?

Mike-3:
1. Why percent instead of real expected output? I am not quite sure, maybe could work also. I was just wondering that you decided % instead of width-px as we chose in the call... Well, I guess, it's ok. Maybe you can add an information line with the expected width/height for chosen scale?

was easier for me to understand. The user does not have to enter the width and height and also calculate them himself so that it is not distorted. 

2.    Default is scale = 0%. Shouldn't it be: 100% (original size). 50% would be half the size and 200% doubled (per side of course). This is how all image tools work.

The default is 100%. I don't see why it doesn't show 100. 

3.    Changing the scale value affects the offsetY value in decrypt. I guess you use the same variable for both inputs. Please correct.

Scale is only used when generating. The specified offset is actually applied to the scaled image.
I can't understand, I look at myself again. 

Furthermore:
I created an output with 50% and one with 100%. Then for tests I used one image from 50 and one for 100. In my expectation one quarter of the resulting image should be darker (the one where the half size overlaps the full size) and the rest should look quite normal. But I got this:

Non-overlapping areas are shown in black. Therefore only a quarter of the picture is b / w.
The areas that do not overlap have no usable result. 

Image Views:
You chose to but the image buttons on the top instead of right. This add extra scroll space. I'd tend to put them right (which is expected to be default in the GCWImageView widget for exactly the space reason). What do you mean?

I change this.

PR #641, opened 2021-06-25
book cipher extension

Mike-3:


S-Man42:
grafik

The "Mode" under "Spaces": Is it "empty lines"? or "ignore symbols"? or...

I'd make Spaces and Empty lines "off" by default. I recently had to deal with several book ciphers... All had spaces and empty lines off (only one, which explicitly mentioned, to recognizes spaces as well).

For "ignore symbols" i'd make a switch as well. So you can disable the ignoring without clearing the entire text field...

S-Man42:
Maybe "only first word letter" not under "advanced" but as general option?

Mike-3:
The "Mode" under "Spaces": Is it "empty lines"? or "ignore symbols"? or...
fixed

So until now I had almost only book cipher, where line-word-letter was given. So if you deactivate spaces, you can no longer determine a word. Then just every line would be a new word.
But I disable it by default.

Mike-3:
book_chipher

ok ?

S-Man42:
Yes, not all combinations make sense at all, of course... I guess that will do the job, I'll test it.

S-Man42:
Do you mind the conflicts?

PR #656, opened 2021-07-30
implement hanja, hangul, sino-korean, korean - https://www.geocaching.com/geocache/GC9ED3B_ein-cache-fur-heidi57i

t-m-z:


lib/logic/tools/crypto_and_encodings/numeral_words.dart
```diff
@@ -1102,6 +1092,239 @@ final Map<String, String> SHAWordToNum = {
...
+  'yeoldul': '12',
+  'yeoldool': '12',
+  'yeoldu': '12',
+  'yeolseht': '13',
+  'yeolset': '13',
+  'yeolse': '13',
+  'yeolneht': '14',
+  'yeolnet': '14',
+  'yeolne': '14',
+  'yeol': '15',
```
S-Man42, line 1250: Are you sure, that 15 to 19 are the same word?

t-m-z: obviously not - copy and paste error - corrected

PR #661, opened 2021-08-02
Tmz 2.0.0 braille graphic simple

t-m-z:
For a first review and discussion.

On the long run I intend to integrate five options for encoding/decoding:
- simple Braille - only A-Z, 0-9
- DEU, ENG, FRA Braille 
- Euro Braille

Regarding 2.0.0 I would recommend only two options:
- simple Braille - only A-Z, 0-9
- Euro Braille


S-Man42:
- Switch from decode to encode: Red Screen of Death
- I tried to enter the symbol for I:
```
.O
O.
..
```
No output on that

t-m-z:
Red Screen of Death was caused by an invalid initialisation - should work now and give an impression.
As I wrote, DEU, ENG and FRA will be an enhancement

S-Man42:
1. Simple Map: F/6 is still noch working on decoding. Please check every character again carefully.
2. Encoding: "ABC" (Capital case) is not working.
3. On "Simple Encoding", maybe switch between Letters and Numbers as well (adding the switch character)?

assets/i18n/en.json
```diff
@@ -347,6 +347,17 @@
   "book_cipher_output_format": "Output Format",
   "book_cipher_input_format": "Positions Format",
   "book_cipher_random_output_hint": "Letters are chosen randomly",
+  "braille_selection_title" : "Braille",
+  "braille_selection_description" : "Encoding/decoding of Braille",
```
S-Man42, line 351: In German title I'd use "Braille (Blindenschrift)" to explain the tool better for those who haven't heard about the term "Braille".

In English, it seems that they have not special word for "Blindenschrift", but they use the term "Tactile Script" in some case. Maybe this could be added in the description somehow.

lib/widgets/registry.dart
```diff
@@ -497,12 +499,29 @@ class Registry {
...
+          ]),
+
+    GCWTool(
+    tool: BeaufortSelection(),
+    i18nPrefix: 'beaufort_selection',
+    categories: [ToolCategory.SCIENCE_AND_TECHNOLOGY],
           helpLocales: ['de', 'en', 'fr'],
-          searchStrings: []),
+          searchStrings: [
+            SEARCHSTRING_COMMON_BEAUFORT,
```
S-Man42, line 521: We removed the Searchstrings from the Selections to avoid confusing results with two similar entries. If you searched for "Beaufort" you got the "Beaufort" Entry for the selection and the "Beaufort" Entry for the actual tool. This came out was more confusing than helpful because now the user did not know which to click - and if he choose, he hit the selection and got the same list displayed he currently saw, because the selection list is nearly the same as the search result.

So, please remove the entries again ;)

lib/widgets/registry.dart
```diff
@@ -497,12 +499,29 @@ class Registry {
...
       GCWTool(
-          tool: BeaufortSelection(),
-          i18nPrefix: 'beaufort_selection',
-          categories: [ToolCategory.SCIENCE_AND_TECHNOLOGY],
+          tool: BrailleSelection(),
+          i18nPrefix: 'braille_selection',
+        categories: [ToolCategory.CRYPTOGRAPHY],
+    helpLocales: ['de', 'en', 'fr'],
+          searchStrings: [
+            SEARCHSTRING_COMMON_BRAILLE,
```
S-Man42, line 509: See Beaufort comment ;) Please leave the search strings in the selection.

lib/widgets/tools/crypto_and_encodings/braille_euro_segment_display.dart
```diff
@@ -0,0 +1,106 @@
+import 'package:flutter/material.dart';
```
S-Man42, line 1: Your graphical segment widget is not centered. You can play around with the RELATIVE_DISPLAY_WIDTH/HEIGHT to get this better. However, I can adjust it for you afterwards if you like :)

t-m-z: I have already noticed that but there is a lack of experience.
I would accept your offer.

lib/widgets/registry.dart
t-m-z: Okay - this was a result of a merge conflict. Done.

lib/widgets/registry.dart
t-m-z: Done

lib/widgets/registry.dart
t-m-z: But there are a still a lot of selections with searchstrings => astronomy, bcd, ... I will ckeck and remove

S-Man42: Yes, it's kind of different... astronomy should be find when typing "astronomy", but then, sun/moon are not added explicitly or something... One have to check the case... Sometimes it makes sense, sometimes not.

BCD same: The specific things should not be listed when typing BCD...

t-m-z:
Thanks for your hints regarding syntax etc. The most interesting hint/question was "3. On "Simple Encoding", maybe switch between Letters and Numbers as well (adding the switch character)?" This is what I am interested in. How should this tool work and be used.
From my experience Braille encoded coordinates used either the alphabet and numeral words or only numbers but no mixture.
So the most simple way would be using only the alphabet with numeral words.
Then switching between letters and digits.
Adding the switch character enhances complexity and would be the third stage.
Adding grammar - Capitals, Math etc. would be the fourth stage.
The next thing: French Braille encodes digits with a different system than German or English.
Over all it is really confusing having only national standards with a minimum common subset of a-z.

I intend to realize 
- Simple Braille A-Z, 0-9 offering for decoding the option to choose between letters and digits. In case of digits choosing the language French or German/English
- Quite Simple Braille with the "number switch"
- Braille with Capitals, Numbers
- Euro-Braille (8 dots)

t-m-z:
Examples - for testing
GC1WPAE Braille https://www.geocaching.com/geocache/GC1WPAE_braille
GC5X6C8 Braille Numbers https://www.geocaching.com/geocache/GC5X6C8_braille-numbers
GC5PZ8Z Braille https://www.geocaching.com/geocache/GC5PZ8Z_braille
GC1N24V Code Braille https://www.geocaching.com/geocache/GC1N24V_code-braille
GC7QK85 (Braille Cube)³ https://www.geocaching.com/geocache/GC7QK85_braille-cube

PR #667, opened 2021-08-07
additional Symboltables: Niessen similar to Lorm and ColorAdd

t-m-z:


assets/i18n/en.json
```diff
@@ -3256,7 +3256,35 @@
...
+  "symboltables_color_add_light_blue" : "Light Blue",
+  "symboltables_color_add_light_orange" : "Light Orange",
+  "symboltables_color_add_light_green" : "Light Green",
+  "symboltables_color_add_light_brown" : "Light Brown",
+  "symboltables_color_add_light_purple" : "Light Purple",
+  "symboltables_color_add_dark_red" : "Dark Red",
+  "symboltables_color_add_dark_yellow" : "Dark Yellow",
+  "symboltables_color_add_dark_blue" : "Dark Blue",
+  "symboltables_color_add_dark_orange" : "Dark Orange",
+  "symboltables_color_add_dark_green" : "Dark Green",
```
S-Man42, line 3285: I think it could make sense to put this in a section "common_colors" (if not yet exists from RAL colors...) What do you think?

lib/widgets/registry.dart
```diff
@@ -5875,6 +5875,12 @@ class Registry {
         SEARCHSTRING_EN_CISTERCIAN,
         SEARCHSTRING_FR_CISTERCIAN
       ]),
+      GCWSymbolTableTool(symbolKey: 'color_add', symbolSearchStrings: [
+        SEARCHSTRING_COMMON_SYMBOL_COLOR_ADD,
+        SEARCHSTRING_DE_SYMBOL_COLOR_ADD,
```
S-Man42, line 5880: Maybe something like COMMON_COLORS here as well?

t-m-z:
Regarding RAL - there are no dublifications.
But Karol Robot could participate from this section.
I will incorporate this.


Sent from MailDroid

-----Original Message-----

t-m-z:
but does the config file with the translation has access to a common color section?
I am not sure/convinced.

Sent from MailDroid

-----Original Message-----

t-m-z:
I think not.
As far as I understood ColorAdd is used to translate a limited set of common colors into a black-white symbology.
You have the three base colors red, blue, yellow which were in added:
red+yello=orange
red+blue=purple
yellow+blue=green
...
the white background means light,
the black background means dark.

Hence I would focus on identification system and not on colors.


Sent from MailDroid

-----Original Message-----

S-Man42:
Regarding common_colors:
Yes, I recently added the opportunity for country_flags. Olease have a look at this to see how it is solved :)

Reg. Search:
I just saw the name and your color string... I think it could make sense to add this to the search when typing "color" or "Farbe". But in the end you should decide. I have no clue what the symbols are. But think of it as a user who has no clue as well, just seeing some colorful symbols...

We have the search string "color" a few times, i guess it makes sense to put this into a common_color string if not yet exists. Dont you think?

t-m-z:
It is amazing how this App evolves. Thanks for this hint. 
Regarding color I will check and adjust.

Sent from MailDroid

-----Original Message-----

t-m-z:
add common_color_section 
add searchstrings color
refactor registry
refactor config.file for color_add

Issue #271, opened 2020-11-30
Piano key frequences


S-Man42:
https://de.wikipedia.org/wiki/Frequenzen_der_gleichstufigen_Stimmung

Birdy2404:
[piano_keys.txt](https://github.com/S-Man42/GCWizard/files/6948711/piano_keys.txt)

t-m-z:
please check with french wikipedia, too:
https://fr.wikipedia.org/wiki/Fr%C3%A9quences_des_touches_du_piano

Birdy2404:
[piano_keys.txt](https://github.com/S-Man42/GCWizard/files/6950682/piano_keys.txt)
Added also the 2 french versions

PR #669, opened 2021-08-09
tmz 2.0.0 Piano key frequences #271

t-m-z:
just for discussion


S-Man42:
Without seeing your work: My intension was doing it exactly like the zodiac tool...

t-m-z:
This answers my question .
As soon as I saw the result of my work I was sure there should be a better solution.

S-Man42:
:D

Sry for being not more constructive... Just wanted to say what was in my mind. I am still not home, couldnt have a look at your solution. So, maybe yours is good or even better - you must decide :)

t-m-z:
No it is okay - as I wrote: I saw my result and was sure, that is hitting the dart board but not bulls eye.
If I had seen your Zodiac earlier I would have chosen this UI. 
So do not waste your time I will come back

t-m-z:
second try    ;-)

PR #671, opened 2021-08-12
Beghilos

Mike-3:


S-Man42:
1. Here it makes totally sense to add the graphical 7 segment output as well.
2. I'd ignore case. So: "bill" = "BILL" = "7718"
3. Decrypt should only accept numbers. So, you should use two separate TextFields with two Controllers. For the Decryption text field you should add a hintText "Enter Numbers".

Mike-3:
changed

PR #675, opened 2021-08-15
Tmz 2.1.0 up-down-left-right e.g. wasd ijkm

t-m-z:
A first preview to get your hints and suggestions


S-Man42:
grafik

* Why is the graphical output not aligned?
* Why are there missing segments (first 4, the 22 in the end)

grafik

In this picture I tried to draw a second figure next to the first one. But with "AAAA" it drawed into the first one instead of shifting itself to right... Is there a way to draw this second figure next to the first one? Maybe through adding more spaces?

```
----|
    |
```

* I'd show both outputs in both modes as in Segment views or whereever. First the graphical interpretation, then the text which was found. In both, the encryption too :)
* The text shouldn't be output with spaces: instead "1 2 3 4 5" -> "12345". If one figure couldnt be recognized, maybe an "UNKNOWN" character (the global variable) should be added?
* The Custom Controls should either a) Allow only one character or b) Did you test, what happens, if the custom controls for WASD: W is not "W" but "WW" or "WASD"? So up is "WASD", Down is "S", ...
* Add controls with arrow symbols, maybe even as the main controls. So, a user mustn't know about gaming controls, but only sees the real "up", "down", "left", "right"... When this mode is chosen, it makes sense for decryption to add a button tool bar like in TomTom or Morse for simpler input. This maybe makes sense for all modes: Just offer 4 buttons, maybe aligned in cursor order ( .:. )
* For Custom Controls: You used a hint text with the arrows. However, the hint text is not visible because you preset the textfield with W, A, S, D. So, it makes more sense to add a text in front of the text field. Because the Textfields should only handle one single character, it could make also sense, to but all four in a row to save space.
* Put the input textfield top, which is more convient to the rest of the tools. And the Controls below, maybe devided by a GCWTextDivider

t-m-z:
- Why is the graphical output not aligned?
=> horizontal spacing is adjusted
=> vertical alignment is under development
- Why are there missing segments (first 4, the 22 in the end)
=> errors in the map - corrected

- In this picture I tried to draw a second figure next to the first one. But with "AAAA" it drawed into the first one instead of shifting itself to right... Is there a way to draw this second figure next to the first one? Maybe through adding more spaces?
=> will be corrected - it is also a question of horizontal spacing/alignment

- I'd show both outputs in both modes as in Segment views or whereever. First the graphical interpretation, then the text which was found. In both, the encryption too :)
=> hence I can delete the Decode mode switch Text/graphic. 

- The text shouldn't be output with spaces: instead "1 2 3 4 5" -> "12345". If one figure couldnt be recognized, maybe an "UNKNOWN" character (the global variable) should be added?
=> done - this is not a big issue
=> UNKNOWN is already implemented

- The Custom Controls should either a) Allow only one character or b) Did you test, what happens, if the custom controls for WASD: W is not "W" but "WW" or "WASD"? So up is "WASD", Down is "S", ...

- Add controls with arrow symbols, maybe even as the main controls. So, a user mustn't know about gaming controls, but only sees the real "up", "down", "left", "right"... When this mode is chosen, it makes sense for decryption to add a button tool bar like in TomTom or Morse for simpler input. This maybe makes sense for all modes: Just offer 4 buttons, maybe aligned in cursor order ( .:. )

- For Custom Controls: You used a hint text with the arrows. However, the hint text is not visible because you preset the textfield with W, A, S, D. So, it makes more sense to add a text in front of the text field. Because the Textfields should only handle one single character, it could make also sense, to but all four in a row to save space.

- Put the input textfield top, which is more convient to the rest of the tools. And the Controls below, maybe devided by a GCWTextDivider
=> yes - however, there are still a lot of tools we have to check and adapt

S-Man42:
I'll merge it. However I'd like to add the Cursor Buttons and we should think about an algorithm for detecting WWSS as 1 as well :)

PR #677, opened 2021-08-15
add new table: GC Attributes and their meaning

t-m-z:


S-Man42:
Did you test this? I didnt so far, but in my opinion the zip files need the name of the directory... Am I wrong?

Mike-3:
This is not necessery

PR #703, opened 2021-08-31
tmz icao addon from https://de.wikipedia.org/wiki/Liste_der_Fliegerhorste_in_Deutschland

t-m-z:


t-m-z:
yes. a hick up due to length of file.

Sent from MailDroid

-----Original Message-----

t-m-z:
a more interetsing question ist - how to deal with air fields with an empty IATA-Code?

S-Man42:
What do you mean? If there's no IATA code, you cannot search it in IATA mode... But you find it in ICAO then...

PR #737, opened 2021-09-18
tmz 2.1.0 Telegraph selection

t-m-z:
first sneak prview


Mike-3:
is this correct with this new files ? Not only in the zip-file ?
for example: assets/symbol_tables/murray_v4_roucoux/0.png

t-m-z:
> is this correct with this new files ? Not only in the zip-file ? for example: assets/symbol_tables/murray_v4_roucoux/0.png

ups - thanks for this hint. of course not.

PR #787, opened 2021-11-02
Multi decoder extension

Mike-3:
features added:
Beghilos
Binary2image
Keyboard layouts
Wasd
Chronogram
Ccitt1/ Ccitt2
Numbers to Braille

add password text field

Playfair
Polybios
Tapir
OneTimePad
Vigenere


S-Man42:
Playfair/Polybios: Two Descriptions found. Instead of "Mode: Modification: ..." it should only be "Modification:"
grafik

S-Man42:
On Keyboard:
grafik
the output set values should be translated, not the internal keys, as they are in the DropDown:
grafik

S-Man42:
What is B/W Image?

S-Man42:
grafik
Mode = true shouldn't be displayed. Similar to Keyboard issue

S-Man42:
grafik
BCD issue. I thought I already fixed this in 2.0.1...
https://github.com/S-Man42/GCWizard/commit/7ddbfc0edbcae125c9926ba32d5a487a4a8c1123

Why is the problem already existing?

S-Man42:
An idea: If key is given, Codes without key don't need to be displayed. If the user searches for a result which uses a key, he is not interested in rotation... And vice versa: Without a given key, vigenere or playfair doesn't make sense

Birdy2404:
Just my opinion: If a user knows, what he is searching for, he will not use the multi decoder 😉
So it could be a usecase that you just type in GC-Code as key to get all the possible results.

Mike-3:
> 
> 
> Playfair/Polybios: Two Descriptions found. Instead of "Mode: Modification: ..." it should only be "Modification:" grafik

I've seen that too, but I don't think there's an easy solution. The widget already has the labeling modification and the display in the tool automatically puts a label in front of the specified widget.

Mike-3:
> 
> 
> On Keyboard: grafik the output set values should be translated, not the internal keys, as they are in the DropDown: grafik

I don't see any possibility of change at the moment. The saved setting is displayed. If we were to save the translated texts of the setting, in my opinion there would be problems if the translation or language changes. Then the setting can no longer be assigned to the internal enum. Even the default value causes problems, for which language should it be saved?

Mike-3:
> 
> 
> What is B/W Image?

Binary to image (Black/ White Image)

Mike-3:
> 
> 
> An idea: If key is given, Codes without key don't need to be displayed. If the user searches for a result which uses a key, he is not interested in rotation... And vice versa: Without a given key, vigenere or playfair doesn't make sense

The problems that come to my mind first. Which text from the website should be evaluated. The whole listing would be too much. I could possibly imagine for the image analysis. And secondly, without user access data, we probably have no way of downloading the page. But we probably 'have to' deposit this data at some point anyway.

Mike-3:
> 
> 
> An idea: If key is given, Codes without key don't need to be displayed. If the user searches for a result which uses a key, he is not interested in rotation... And vice versa: Without a given key, vigenere or playfair doesn't make sense

At the moment, only those tools are hidden that require a password and none is specified (result == null).
But I change that the tools without a password, when specifying a password, do not return anything and are thus hidden.

Mike-3:
> 
> 
> grafik BCD issue. I thought I already fixed this in 2.0.1... [7ddbfc0](https://github.com/S-Man42/GCWizard/commit/7ddbfc0edbcae125c9926ba32d5a487a4a8c1123)
> 
> Why is the problem already existing?



> 
> 
> grafik BCD issue. I thought I already fixed this in 2.0.1... [7ddbfc0](https://github.com/S-Man42/GCWizard/commit/7ddbfc0edbcae125c9926ba32d5a487a4a8c1123)
> 
> Why is the problem already existing?

Also changed in brunch 2.1? I have no idea why the problem is still there. Merge problem? I correct it.

S-Man42:
> 
> 
> > Playfair/Polybios: Two Descriptions found. Instead of "Mode: Modification: ..." it should only be "Modification:" grafik
> 
> I've seen that too, but I don't think there's an easy solution. The widget already has the labeling modification and the display in the tool automatically puts a label in front of the specified widget.

Hmmm.... maybe we should add a "suppressTitle" in the widget and then we can put the label here...

S-Man42:
> 
> 
> > An idea: If key is given, Codes without key don't need to be displayed. If the user searches for a result which uses a key, he is not interested in rotation... And vice versa: Without a given key, vigenere or playfair doesn't make sense
> 
> The problems that come to my mind first. Which text from the website should be evaluated. The whole listing would be too much. I could possibly imagine for the image analysis. And secondly, without user access data, we probably have no way of downloading the page. But we probably 'have to' deposit this data at some point anyway.

What are you talking? :D Somehow question and answer doesnt fit here... Or am I wrong? :D

S-Man42:
> 
> 
> > An idea: If key is given, Codes without key don't need to be displayed. If the user searches for a result which uses a key, he is not interested in rotation... And vice versa: Without a given key, vigenere or playfair doesn't make sense
> 
> At the moment, only those tools are hidden that require a password and none is specified (result == null). But I change that the tools without a password, when specifying a password, do not return anything and are thus hidden.

I am sure that this might be very useful. We want to support the user, not give him results which are not useful. Giving results for password related ciphers without a password are not useful as well as giving non-password related cipher using a password (latter a bit less, because it may be that the user found a password which is none.).

Password-related ciphers usually give some output because we designed them to do so - working with edge cases. But in this use case, they just make the list much longer than it needs to be. 

@Birdy2404 And if the user really clicks "Advanced" and then enters the ciphertext into key field and leaves the normal input field clear, than... sorry, than comes the rule: Trash in, trash out. At some point we need to believe there's still some brain. Every tool can be used the wrong way. We cannot ensure for every wrong usage a meaningful result. Here as well...

Mike-3:
> 
> 
> > > Playfair/Polybios: Two Descriptions found. Instead of "Mode: Modification: ..." it should only be "Modification:" grafik
> > 
> > 
> > I've seen that too, but I don't think there's an easy solution. The widget already has the labeling modification and the display in the tool automatically puts a label in front of the specified widget.
> 
> Hmmm.... maybe we should add a "suppressTitle" in the widget and then we can put the label here...

I think this is possible.

S-Man42:
> 
> 
> > grafik BCD issue. I thought I already fixed this in 2.0.1... [7ddbfc0](https://github.com/S-Man42/GCWizard/commit/7ddbfc0edbcae125c9926ba32d5a487a4a8c1123)
> > Why is the problem already existing?
> 
> > grafik BCD issue. I thought I already fixed this in 2.0.1... [7ddbfc0](https://github.com/S-Man42/GCWizard/commit/7ddbfc0edbcae125c9926ba32d5a487a4a8c1123)
> > Why is the problem already existing?
> 
> Also changed in brunch 2.1? I have no idea why the problem is still there. Merge problem? I correct it.

Yes, I currently had a look on this. It seems that "_title" is still present in https://github.com/S-Man42/GCWizard/commit/7ddbfc0edbcae125c9926ba32d5a487a4a8c1123#diff-9ab19e3a4995a8856ec593ff24114e89716a05e226faa4761cbe535e071503e6L53

Mike-3:
> 
> 
> > > An idea: If key is given, Codes without key don't need to be displayed. If the user searches for a result which uses a key, he is not interested in rotation... And vice versa: Without a given key, vigenere or playfair doesn't make sense
> > 
> > 
> > At the moment, only those tools are hidden that require a password and none is specified (result == null). But I change that the tools without a password, when specifying a password, do not return anything and are thus hidden.
> 
> I am sure that this might be very useful. We want to support the user, not give him results which are not useful. Giving results for password related ciphers without a password are not useful as well as giving non-password related cipher using a password (latter a bit less, because it may be that the user found a password which is none.).
> 
> Password-related ciphers usually give some output because we designed them to do so - working with edge cases. But in this use case, they just make the list much longer than it needs to be.
> 
> @Birdy2404 And if the user really clicks "Advanced" and then enters the ciphertext into key field and leaves the normal input field clear, than... sorry, than comes the rule: Trash in, trash out. At some point we need to believe there's still some brain. Every tool can be used the wrong way. We cannot ensure for every wrong usage a meaningful result. Here as well...

So the ttols that need a password are hidden, if none was given is already like that. I think Vigenere has an exception that I can change).
The tools without a password are hidden if a password has been specified.

Birdy2404:
> @Birdy2404 And if the user really clicks "Advanced" and then enters the ciphertext into key field and leaves the normal input field clear, than... sorry, than comes the rule: Trash in, trash out. At some point we need to believe there's still some brain. Every tool can be used the wrong way. We cannot ensure for every wrong usage a meaningful result. Here as well...


I don't think in missuse. I think in the usecase that I want all results and therefore try also the GC code in the key field.

S-Man42:
What should be the use case for that? You got a GCCode and put it into the key field without any ciphertext entered? What should this be?

Or do you mean: Everything can be a password, I put the GCCode everywhere, just in case... (like my comment above: User found a "password" which is none... in that case the GCCode is no password). Well in that case the user can try twice: Once without any password and gets the non-password ciphers and than with password to get the other ones... It's ok for me to let the user do several tries. The pro of this is, that in big majority of all cases the resulting list makes much more sense.

PR #789, opened 2021-11-05
tmz flags royal navy: howe, popham, marryat

t-m-z:
add three symbol tables with signalling flags of the royal navy


S-Man42:
missing searchstrings

PR #794, opened 2021-11-15
add formula colors (optional)

Mike-3:
with tests


Mike-3:
text is selectable

colors:
text -> standard color
variable, constant -> orange (constant also orange, good idea ?)
functions, brackets -> blue
formula references -> blue
numbers -> green
errors -> bold red

PR #797, opened 2021-11-17
Tmz 2.1.0 predator

t-m-z:
add predator code
- symbol table
- graphical en/decryption


S-Man42:
Please resolve conflicts

S-Man42:
Please add search to symbol tables. Please remove it from selection. (When searching, you want to see the real tools, not another menu, we tried to remove the searchlists where it makes sense)

The graphical encoding... I am not quite happy with the result. Some of the bars are very hard to click (too slim) and the entire widget looks kinda stretched... Maybe there one could improve a bit?

assets/searchstrings/common.json
```diff
@@ -283,6 +283,7 @@
   "primes_nearestprime": "",
   "primes_nthprime": "",
   "primes_primeindex": "index",
+  "predator": "predator",
```
S-Man42, line 286: predators

t-m-z:
> Please resolve conflicts

done

assets/searchstrings/common.json
t-m-z: I do not understand this. According to google it is predator.

t-m-z:
> Please add search to symbol tables. Please remove it from selection. (When searching, you want to see the real tools, not another menu, we tried to remove the searchlists where it makes sense)
> 
> The graphical encoding... I am not quite happy with the result. Some of the bars are very hard to click (too slim) and the entire widget looks kinda stretched... Maybe there one could improve a bit?

done and tested on my phone. seems to be okay

assets/searchstrings/common.json
S-Man42: haha, yes, but with predators you can also search for many of these wonderful guys ;)

t-m-z: done

assets/searchstrings/common.json
```diff
@@ -480,6 +481,7 @@
   "symbol_postcode8421": "postal service 8421 4",
   "symbol_postnet": "postnet united states postal service usps",
   "symbol_puzzle": "",
+  "symbol_predator": "predator",
```
S-Man42, line 484: why didn't you change here as well? maybe it makes sense to use only one of the searchstrings. So, take "predator" instead of "symbol_predator" for the symbol table as well to remove duplicates

t-m-z: to be honest - it was a little bit late.
But it was agood advice to reduce the duplicates

PR #809, opened 2021-11-28
tmz 2.1.0 Telegraph section

t-m-z:
add a new section to crypto and encodings for telegraphs 
- Chappe (France)
- Edelcrantz (Sweden)
- Murray (England)
- Ohlsen (Norway)
- Prussia
- Semaphore
- Wig Wag
- Gauss Weber
- Schilling Canstatt
- Wheatsone Cooke
- Punch Tape (Baudot/Murray)

correct Symbol Tables

Only one MAJOR Problem: I have no clue for the correct size of the punch tape - need help
Only on MINOR Problem: One flag at Semaphore is a little bit tricky to tap on


S-Man42:
Phew... What a great work! So many kudos for such effort!

However, there are several issues I identified at first reviews:

Generally: 
1. Your "always black" parts should be better colored. You'll never see is on dark smartphones. What about using "themeColors().mainFont()" for example. This would automatically make the black parts white on dark theme and black on white theme
2. Export colors You should make the "black" parts grey or something on export. Some telegraphs are badly visible with black on black (the telegraph arm directly on a black part)

Specifically:
grafik
Display output scales not correctly

grafik
Markered flag is not clickable, only the... griff...

grafik
Are you quite sure, one could click such detailed parts with its finger on a smartphone?

grafik
No idea if it's correct; but was not able to generate output here

grafik
NULL Pointer

grafik
1. Shouldn't it be centered?
2. Why is there wider black border at the right side? Intended?
3. At day mode, no activated field yield empty result, on night mode not. Empty results shouldn't work anyway...

t-m-z:
Thanks for the first impressions
- regarding colors: I did change the black part to grey. Now they are visible in both modes and also suitable for export
- punchtape - corrected
- the semaphore flag is now fully touchable
- regarding Chappe - It is tricky, but possible to touch these segments
- regarding Ohlsen - this is okay. YXou can only encode according to the codebook. Ohlsens telegraph was a messaging system with fixed messages. The shutter were decoded to numbers which were translated into messages and vice versa.
- WigWag corrected
- regarding Edelcrantz
   1. no-  according to my reasearches
   2. yes - but it is nothinmh important and can be reduced
   3. corrected

S-Man42:
I merged this, because I want to see it in the test version on the smartphone. Maybe some issues may follow :)

PR #811, opened 2021-11-29
Interpolate formula

### LINKED ISSUE #115: Formula Solver: Add function for cross totals
Main question is: How to add custom function to the parser?

### LINKED ISSUE #579: Operation formula solver - variable coordinates
Doesn't it make sense to transfer the operation and functionalities of the form solver to the variable coordinate as well?
(Context menu -> export, import, modify)

### LINKED ISSUE #666: VarCoords: Handle all formulas, even if result is no coordinate
Sometime a formula is not a coordinate but a code for a number key lock or something. And then I also want to display all possibilities...

S-Man42:
* Formula Interpolate
* Custom number functions (cs, csi, ...)
* Custom text functions (bww, av, len)

Description of functions:
* Formula Values have now next to key/value a third type option: 
  * Value: Normal value: Normal number or formula incl. references to other values as currently known
  * Interpolation: As known from VarCoords, used for interpolation
  * Text: Simple Text input, used for text functions
* if interpolation value used and more than one result returned (max 100): 
  * Instead of normal result, an expandable widget appears which shows all solutions and the related variable constellation
  * each interpolation has its own function popup button, incl. map view if coordinate detected
  * three different views for map: only the one interpolation (on interpolation button), all interpolations for the formula (formula popup menu), all coords from all formulas (top popup button.)
* Labels in map should contain {$NumberOfFormula}, if interpolation value {$NumberOfFormula.NumberOfInterpolation}. Interpolated map results are displayed grey instead black

closes #666
closes #579 
closes #115

Please check carefully. It's been a crazy huge change with many many cross and siding effects through the entire app, whereas:
- Hash Breaker (Value Entries)
- Var Coords (entire tool, mainly value entries)
- Alphabet Values (Expert: Value Entries)
- other Value Entry using tools
- several bugfixes in different tool I found on-the-fly

Additionally there is a fix for 3 Circle Intersect: The copy button for the results did copy the additional accurancy information, not only the coords. It may be, that there is also some siding effect in the coords


Mike-3:
solve confilict ;-)
first notes:
When you select edit formula, the focus always jumps to the field to be added and not to the field to be edited. But it was probably like that before. 
Perhaps toggle through the type by tapping it directly. So you don't always have to put the field into editing mode first.

Mike-3:
If formula and numbers are mixed, e.g. (9 + 3) + test, the expression in brackets is not calculated. Can that still be done?

S-Man42:
"When you select edit formula ,... "
- I will check, but cannot promise

"Perhaps toggle through the type by tapping it directly"
- I had it first. However, just showing symbols without text explanation seems very hard for the users. In fact, only crazy advanced people will use other options than the first one. And this kindly rare, I believe.

"If formula and numbers are mixed, "
- simply no. It didn't work so far and it won't in future. Reason: Entire term is given into the mathematical engine. For the engine this is an invalid term. To solve the brace term beforehand would yield an own equation parser which is much too crazy.

S-Man42:
Fixed conflict and (maybe) focus issue (although it has been there since beginning)

Mike-3:
grafik
grafik

I found the old relationship (bottom) better. Can you change that for the other tools outside of the formula sover?

S-Man42:
Done.

Mike-3:
review finished.
no problems found

PR #816, opened 2021-12-07
qr code upgrade

Mike-3:
version 3.0
integer spinner modul size


assets/i18n/en.json
```diff
@@ -2863,6 +2863,7 @@
   "qr_code_title": "QR Code",
   "qr_code_description": "Recognize QR code from image files\nCreate QR codes",
   "qr_code_nothingfound": "No QR code found.",
+  "qr_code_modulsize": "Modul size",
```
S-Man42, line 2866: Maybe we should call it "Pixel Size".... A user doesnt know what a module is...

PR #825, opened 2021-12-19
Multi decoder roman numbers bugfix

Mike-3:


S-Man42:
chronogram has the same issue...

PR #880, opened 2022-01-17
Implement additon symbol tables

Mike-3:


S-Man42:
1. Please always check licenses. E.g. in "Face It", license is only "Free for personal use" (as it can seen in the "Codeschnipsel" file, @Birdy2404 added). In case it is not completely free (also for commercial purposes), we need to attribute them seperately in these files (otherwise, we ):
* https://github.com/S-Man42/GCWizard/blob/2.2.0/LICENSE
* https://github.com/S-Man42/GCWizard/blob/2.2.0/lib/widgets/main_menu/licenses.dart

2. Please do not deliver German translations in commit. They may cause conflicts to the Crowdin translations. Only place for translations is Crowdin. Furthermore I don't like the description texts because they differ from the current schema: "From the video game 'Zelda: Blakeks'" Should be aligned (@Birdy2404)

3. Please introduce a new common searchstring (common "zelda": "zelda hyrule whatever"; language specific "zelda": "hylians" or "hylianisches" or whatever may be) for all related symbol tables (the many new ones as well as the two or three already existing). Remove "zelda" and similar terms from specific symbol search strings and add new "zelda" to all those instead.

PR #893, opened 2022-01-22
renamed genreich -> gernreich

Mike-3:


S-Man42:
Hmmm, when I search for Genreich, famous GC pages show me this symbols using this name. Are you quite sure about this change? What is the source?

Furthermore it should be added to the search strings (in case

PR #900, opened 2022-01-24
arcadian transferred to common and greek deleted

Birdy2404:


S-Man42:
No, because the sole "skies of" has no meaning anymore. If there comes a translater, he will translate the phrase. Althought there is no title in his language for this game. Either nothing in en (and it will never be translated) and "skiesofarcadian" into common or "skiesofarcadian" into en, so there's a chance to translate it into some  languages. Now I have no knowledge whether this title is commonly translated or not. As far as I can see, it is a common string, because in Wikipedia all languages to the article keep the english title

PR #911, opened 2022-02-06
tmz 2.2.0 base58 base91 base122

t-m-z:
ready for review


lib/widgets/main_menu/licenses.dart
```diff
@@ -78,6 +78,10 @@ class LicensesState extends State<Licenses> {
           ['Urwigo Tools', '@Krevo (WherigoTools)', 'MIT License'],
           ['Vigenère Breaker', 'Jens Guballa (guballa.de)', null],
           ['Whitespace Interpreter', 'Adam Papenhausen', 'MIT License'],
+          ['Base58', 'Dark Launch\nhttps://www.darklaunch.com/base58-encode-and-decode-using-php-with-example-base58-encode-base58-decode.html', ''],
```
S-Man42, line 81: Please remove the urls from here. It makes the licenses view less readable. Putting the urls to the code parts is enough, I believe.

Then you can do:

name1\n
name2

license1\n
license2

:)

lib/widgets/registry.dart
```diff
@@ -1140,6 +1147,14 @@ initializeRegistry(BuildContext context) {
       'base',
       'base85',
     ]),
+    GCWTool(tool: Base91(), i18nPrefix: 'base_base91', searchKeys: [
+      'base',
+      'base122',
```
S-Man42, line 1152: Wrong search string

lib/widgets/main_menu/licenses.dart
t-m-z: check and done

lib/widgets/registry.dart
t-m-z: ups corrected

PR #914, opened 2022-02-07
tmz 2.2.0 wig analyze performant

t-m-z:
The tool seems to work - however it is not easy to deal with LUA Sourcecode buidl with different IDEs.
Hence surprises could arise


assets/i18n/en.json
```diff
@@ -4924,6 +4933,167 @@
...
+  "wherigo_error_gwc_header": "arose while reading Cartridge Header",
+  "wherigo_error_gwc_luabytecode": "arose while reading Cartridge LUA Byte Code",
+  "wherigo_error_gwc_mediafiles": "arose while reading Cartridge Media Files",
+  "wherigo_error_lua_line": "around Line number",
+  "wherigo_error_lua_characters": "arose while reading Character Objects from LUA Sourcecode",
+  "wherigo_error_lua_items": "arose while reading Item Objects from LUA Sourcecode",
+  "wherigo_error_lua_tasks": "arose while reading Task Objects from LUA Sourcecode",
+  "wherigo_error_lua_inputs": "arose while reading Input Objects from LUA Sourcecode",
+  "wherigo_error_lua_answers": "arose while reading answers from LUA Sourcecode",
+  "wherigo_error_lua_zones": "arose while reading Zone Objects from LUA Sourcecode",
```
S-Man42, line 5091: better word in that contexts is "raised". It's common to "raise an exception" :)

lib/logic/tools/crypto_and_encodings/wherigo_urwigo/krevo.wherigotools/readestring.dart
```diff
@@ -0,0 +1,79 @@
...
+        c = x;
+    }
+    result = result + c;
+  }
+  result = result.replaceAll('\004', '&lt;').replaceAll('\005', '&gt;').replaceAll('\006', '&amp;');
+  return result;
+}
+
+
+/**
```
S-Man42, line 39: please remove unneccessary code

assets/i18n/en.json
```diff
@@ -4751,7 +4761,6 @@
   "wasd_custom_down" : "Down",
   "wasd_custom_right" : "Right",
   "wasd_custom_left" : "Left",
-  "wherigourwigo_selection_title": "Wherigo / Urwigo Tools",
```
S-Man42, line 4754: Why did you remove this? I guess the selection needs the text in english...

S-Man42:
Looks good at the first glance! :)

Some issues:

1. When loading a LUA Code not fitting to the GWC file, then it might cause crashes:
(example: s-man42.de/2.gwc with s-man42.de/1.lua -> Show List of Media Files and click through the list. Item 3 will show you red screen of death.) To fix that maybe you should add a check after loading a gwc and lua file to make it plausible they fit (same numbers of media files or whatever is possible... maybe you could check the lua code for some hints to the fitting gwc file: name, bytes, whatever)

2. I find the "GWC-File: Cartiridge" section confusing: I guess some rows are headlines, not data. Please don't use the multiColumnWidget here, it's not clear what it means. What about (quick PowerPoint Mockup):
grafik

--> 1: Not expandable section header
--> 2: Expandable header of subsection
--> 3: For every subsection a separate columnedMultilineDingens with own header line
--> 4: (Just for better UX): columnedMultilineDinges has an option "hasHeaders" or something like this. You should use it here, as it removes the copy button from the header line where it makes no sense at all

Another thought: You use decimal values for offsets. More common is usage of hex values. Don't know whether it makes more sense. Maybe one could add a toggle button to the section header or something. (GCWTextDivider(trailing: GCWIconButton....))

3. For example in "Messages/Dialogs" you simply use a GCWText(), I guess. I was missing the copy button. Maybe instead of GCWText you could use the GCWOutputText() which adds the common copy button. Just to make these section copyable as well :)

S-Man42:
grafik
Please use language specific date format. (See Exif or Changelog for examples :) )

S-Man42:
Sound widget could be enhanced:

When clicking on Play, the Icon should be toggled to Pause. When sound is finished, it can be Play Button again. That gives you the ability to stop the sound when you want.

Next step is to get the entire SoundWidget working, that might help a lot. Some sounds are quite long, so it would be great to jump to a specific point (Example is Mighla, where we recorded the sound from the WIG Player to be able to use it):
grafik

S-Man42:
What's the difference between "Media Files" and "Media Object". Can you merge both views or are they completely different?

S-Man42:
I REALLY DO LOVE IT! Great work!

assets/i18n/en.json
t-m-z: check - done

lib/logic/tools/crypto_and_encodings/wherigo_urwigo/krevo.wherigotools/readestring.dart
t-m-z: done

t-m-z:
- Timer-Objects: Duration: unit (s) added
- Media/media objects are combined
- Hyphens corrected
- ByteCode structure corrected
- Check for different GWC/LUA and react in a proper way
- Messages are now showing the copy-button

- reagarding teh sound player - to be honest; this will take a while

S-Man42:
Concerning the workflow: I would put the dialog not as "first" attention. It may terrify the user, if a long text appears after clicking "WIG analyzer". :D My proposal for the workflow would be:

grafik

As already written via Threema: 
Properly handle http errors. I am not quite sure, what you do? Are you using our file loading engine? (which is used by GCWOpenFile)? There error handling takes place.

S-Man42:
After opening GWC, immediately a "LUA fits not GWC file" message raises. Although no LUA code is currently loaded

S-Man42:
> grafik Please use language specific date format. (See Exif or Changelog for examples :) )

Not implemented?

S-Man42:
Cartridge Structure View:

1. Overall "Structure" Divide not expandable.
2. The subcategories expandable (already is implement) but start collapsed (expanded: false)
3. 
grafik
Remove "Header" row. Headline is second one :)

S-Man42:
Sound ... I have problems with starting sound... when simply press "play", nothing happens... then I wait. nothing. then I switch to next sound. Then I press "play" and get the previous sound. Then I go back and get the correct sound (same as I got with the next). Then I go the next again and get the correct sound there... 

I guess, buffering or loading is not properly handled... I guess some async action must be done here.

assets/i18n/en.json
```diff
@@ -4970,6 +4970,8 @@
   "weathersymbols_ww_99": "Heavy thunderstorm with hail at time of observation",
   "wherigo_title" : "Wherigo Analyzer",
   "wherigo_description" : "Shows Cartridge Data, Media Files, etc. of a Wherigo Cartridge",
+  "wherigo_decompile_title": "IMPORTANT - How to decompile ...",
+  "wherigo_decompile_message": "For a complete analysis of a cartridge the LUA Sourcecode has to be decompiled. For the time beeing this has to be done Online or Offline by youself.\nOnline means, the GC Wizard sends your chosen cartridge - and only the cartridge - nothing else - to our server, which decompiles and sends back the LUA Sourcefile. Otherwise you have to decompile offline with UNLUAC.JAR from Sourceforge.net.\n\nPress OK\n...to agree for Online-Decompiling\nPress CANCEL\n... to decompile on your own",
```
S-Man42, line 4987: Unfortunately the text has many mistakes. Maybe better:

"Unfortunately, the GC Wizard cannot extract the LUA code offline at the moment. If you like, the cartridge (and nothing else!) will be send to our server for analysis.

Alternatively you can put manually decompiled LUA code directly into the tool from a file or an internet resource.

Meanwhile we are searching for an offline solution.

Do you want to proceed with online analysis?"

t-m-z: DateTime surrounded by format-Method like in exif_reader.
Dialog-Text adopted.
GCWStructure refactored.
Check and corrected "After opening GWC, immediately a "LUA fits not GWC file" message raises. Although no LUA code is currently loaded"

t-m-z:
I cannot reproduce your sound problem. e.g. Agent Kar Kartoffel works properly

t-m-z:
add a procentage view on the progress indicator should be possible, but up to now I did not manage to implement it.

t-m-z:
regarding the workflow: I see your point, but this means a deep dive into refactoring

S-Man42:
> I cannot reproduce your sound problem. e.g. Agent Kar Kartoffel works properly

Try s-man42.de/4.gwc. Sounds are media files 152ff.

t-m-z:
STill no Problem within the emulator. I loaded 4.gwc which seems to be mighla_part1 and all media files from 149ff were played without any problem

S-Man42:
* Loaded s-man42.de/4.gwc
* loaded LUA online 
* List possible answers

grafik

S-Man42:
grafik

Example Mighla: The character coordinates seem to be wrong. Definitely not the Mighla playground (which is around Müggelsee):

grafik

S-Man42:
Is this "Callback" thing intended to be shown?

grafik

S-Man42:
Mighla: Dialog/Messages: Dialog 4 and 5 are the same. Problem of the cartridge or bug?
grafik

S-Man42:
grafik
Problem with encoding on our side?

S-Man42:
Please add unit. Either "5 s" or "Duration in s"
grafik

S-Man42:
Isn't the dtable something unordered or am I wrong. Expected something more like a key/password:

grafik

S-Man42:
s-man42.de/2.gwc

I doubt that this is a correct answer ;)

grafik

S-Man42:
Hmm,

grafik

S-Man42:
Congrats, the online decompiler works like a charm. In my few examples I got no problems with it. Very well done!

S-Man42:
I will merge this current version to be able to do the translations meanwhile. You can still use the branch and create a new PR on it when fixes the issues. I will create a test version with the current code, so we can search for bug more efficiently on the devices, ok?

t-m-z:
only Urwigo uses an unordered dTable.

Sent from MailDroid

-----Original Message-----

t-m-z:
however these are the initial coordinates belonging to the character.


Sent from MailDroid

-----Original Message-----

t-m-z:
I know, however the percentage is calculated with regard to the for loop iterating through the LUA source file

Sent from MailDroid

-----Original Message-----

t-m-z:
Thanks - I will pass it to my son   😀

Sent from MailDroid

-----Original Message-----

PR #917, opened 2022-02-11
tmz 2.2.0 numeral words enhancement

t-m-z:
add symbols for
- chinese
- thai
- vietnamese
- japan

I have one problem:
- the symbols for vietnamese chu nom are shown in the dart file but not in the widget. I have no clue what happens with the encoding


lib/logic/tools/crypto_and_encodings/numeral_words.dart
```diff
@@ -1004,7 +1015,137 @@ final Map<String, String> CHIWordToNum = {
...
+  '佰': '100',
+  '仟': '1000',
+  '萬': '10000'
+};
+final Map<String, String> CHIZhuyinWordToNum = {
+// https://de.wikipedia.org/wiki/Chinesische_Zahlzeichen
+  'ㄌㄧㄥˊ': '0',
+  'ㄧ': '1',
+  'ㄦˋ': '2',
+  ' ㄌㄧㄤˇ': '2',
```
S-Man42, line 1120: leading space ok?

lib/logic/tools/crypto_and_encodings/numeral_words.dart
```diff
@@ -1110,7 +1342,76 @@ final Map<String, String> JAPWordToNum = {
...
+  'よん': '4',
+  'ご': '5',
+  'ろく': '6',
+  'しち': '7',
+  'なな': '7',
+  'はち': '8',
+  'きゅう': '9',
+  'く': '9',
+  'じゅう': '10',
+  'じゅういち	': '11',
```
S-Man42, line 1390: ...

S-Man42:
Did you check against the Geocache?

S-Man42:
GC8VBB7

lib/logic/tools/crypto_and_encodings/numeral_words.dart
t-m-z: no - tired eyes problem

lib/logic/tools/crypto_and_encodings/numeral_words.dart
```diff
@@ -1110,7 +1342,76 @@ final Map<String, String> JAPWordToNum = {
   'nana': '7',
   'hachi': '8',
   'kyū': '9',
+  'ku': '9',
   'jū': '10',
+  'jūichi	': '11',
```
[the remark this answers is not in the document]
t-m-z: no - tired eyes problem

lib/logic/tools/crypto_and_encodings/numeral_words.dart
```diff
@@ -1110,7 +1342,76 @@ final Map<String, String> JAPWordToNum = {
...
+  'じゅうに': '12',
+  'じゅうさん': '13',
+  'じゅうよん': '14',
+  'じゅうし': '14',
+  'じゅうご': '15',
+  'じゅうろく': '16',
+  'じゅうしち': '17',
+  'じゅうなな': '17',
+  'じゅうはち': '18',
+  'じゅうきゅう': '19',
```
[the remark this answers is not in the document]
t-m-z: regarding the webpage - yes

t-m-z:
https://www.geocaching.com/geocache/GC8VBB7_biesdorfer-kiste with the two pictures at
- https://img.geocaching.com/cache/thumb/99ace1f8-ccee-4ad0-9e42-c398af76136c.jpeg
- https://img.geocaching.com/cache/thumb/5d5a5e93-b9b4-4571-b07e-0eae6a1cc3dc.jpeg
is something for the symbol replacer tool from Mike

t-m-z:
Hm - I have a Problem with Vietnamese Chu nom:
It seems, that the Roboto Condensed Font does not support these unicode chars:

https://github.com/googlefonts/roboto
The font family supports all Latin, Cyrillic, and Greek characters in Unicode 7.0, as well as the currency symbol for the Georgian lari, to be published in Unicode 8.0.

Do you have an idea?

lib/widgets/tools/crypto_and_encodings/numeral_words/numeral_words_lists.dart
```diff
@@ -70,9 +84,14 @@ class NumeralWordsListsState extends State<NumeralWordsLists> {
...
           context,
           numeralWordsOverview.entries.map((entry) {
-            if (int.tryParse(entry.value) != null) return [entry.value, entry.key];
+            if (int.tryParse(entry.value) != null) {
+              return [entry.value, entry.key];
+            };
           }).toList(),
-          flexValues: [1, 3]),
+          flexValues: [1, 3],
+          fontSize: _valueFontsize * 1.0
```
S-Man42, line 92: * 1? which sense?

lib/widgets/tools/crypto_and_encodings/numeral_words/numeral_words_lists.dart
```diff
@@ -42,7 +46,17 @@ class NumeralWordsListsState extends State<NumeralWordsLists> {
 
     return Column(
       children: <Widget>[
-        GCWDropDownButton(
+        GCWIntegerSpinner(
```
S-Man42, line 49: In general, this is a good idea. However I wouldnt add a separate line for I would put a GCWZoomButtons to the GCWDefaultOutput (trailing) like in the Symboltables. And I believe we should only do this for non-latin symbols.

So

GCWDefaultOutput(
   trailing: If ([CHINESE, THAI,....].contains(_currentLang)
           ? GCWZoomButtons() 
           : null
)

S-Man42:
grafik

When searching for "Zahlwörter" it is not necessary to show the selection. Because its content is already displayed :)

Thai and Thai (Transcription) is switched. The transcription is the words in our symbols. The foreign characters are transcripted into ours :)

Thai and Vietnamese Numbers as symbol tables? Makes this sense?

t-m-z:
The NumeralWordsSelection has an empty list of searchKeys. I have no clue why the selection is displayed.

t-m-z:
Thai and Vietnamese Numbers as symbol tables? Makes this sense?
Yes - because numeral words are not digits. So it is like Tamil numbers

lib/widgets/tools/crypto_and_encodings/numeral_words/numeral_words_lists.dart
t-m-z: convert int to double

S-Man42:
see my comment in morse. i guess it's a verschlimmbesserung

lib/widgets/tools/crypto_and_encodings/numeral_words/numeral_words_lists.dart
S-Man42: well, than make valueFont.toDouble() instead ;)

t-m-z:
okay - so the tool is read for merge from my point of view

PR #922, opened 2022-02-13
tmz 2.2.0 teletypewriter

t-m-z:
refactor ccitt to restore ccitt1, ccitt2
add ccir 476
ad zc1 from AEG Telefunken TR440


S-Man42:
does z22 still work? :)

PR #923, opened 2022-02-13
tmz 2.2.0 symbol morse

t-m-z:
add table


S-Man42:
Maybe it makes sense to add a selection_morse here and adjust the searchstrings

S-Man42:
Searchstring are good. But a selection could make sense nonetheless. Btw. why are "Brainfk" and "Ook" part of the result list when searching for "morse"? Maybe remove?

t-m-z:
> Btw. why are "Brainfk" and "Ook" part of the result list when searching for "morse"? Maybe remove?

No - because we have a derivat named Morsef*ck

t-m-z:
add selection - but I am lost.
Also the selection has no searchKey, it will be shown when searching Morse - I mixed something, but what?

S-Man42:
aaaaah that could be my own fault... i enhanced the search: next to normal searchnstring, it checks the titles of a tool (idea is to provide search support even if the search strings are not yet translated)... aaargh.. i guess i need to dive into this.

PR #933, opened 2022-02-20
TMZ 2.2.0 add Donal Knuth's Potrzebie system of weights and measures

t-m-z:
also fixes a bug


lib/logic/common/units/length.dart
```diff
@@ -59,6 +59,8 @@ final LENGTH_ROD = Length(name: 'common_unit_length_rod_name', symbol: 'rd', inM
 
 final LENGTH_LINK = Length(name: 'common_unit_length_link_name', symbol: 'li', inMeters: 0.201168);
 
+final LENGTH_POTRZEBIE = Length(name: 'common_unit_length_potrzebie_name', symbol: 'p', inMeters: 0.0022633);
```
S-Man42, line 62: please do not round if not neccessary. The internal numeral does it for you, but you shouldnt force it 2.263348517438173216473 ;) And if someone may come and wants to know the values and looks at this code, he believes. Additionally maybe somewhere a link to the website makes sense ;)

t-m-z: ;-)   done

lib/logic/common/units/length.dart
```diff
@@ -80,17 +83,22 @@ final List<Unit> baseLengths = [
...
   LENGTH_PARSEC,
   LENGTH_LAR,
-  LENGTH_RANGAR
+  LENGTH_RANGAR,
+  LENGTH_POTRZEBIE,
 ];
 
 final LENGTH_KM = Length(name: 'common_unit_length_km_name', symbol: 'km', inMeters: 1000.0);
 final LENGTH_CM = Length(name: 'common_unit_length_cm_name', symbol: 'cm', inMeters: 0.01);
+final LENGTH_FARP = Length(name: 'common_unit_length_farshimmelt_name', symbol: 'fp', inMeters: 0.000000022633);
```
S-Man42, line 92: why not updated all occurences? ;)

btw. it maybe makes sense to write 

2.2263.... / 1000.0 or /1000000 or whatever to ensure, that precision will be kept (don't know internally safe system but in Java this makes a difference). Furthermore for a future programmer, he can see, da FARP is calculated from PRODINGENS when there's 2.2 / 1000 / 1000

so best could be:

final LENGTH_POTRZEBIE = Length(name: 'common_unit_length_potrzebie_name', symbol: 'p', inMeters: 2.263348517438173216473 / 1000.0);

final LENGTH_FARP = Length(name: 'common_unit_length_farshimmelt_name', symbol: 'fp', inMeters: 2.263348517438173216473 / 1000.0 / 10000.0); // if this is correct numbers...

t-m-z: oh - that was the moment my called to have a coffee break

PR #939, opened 2022-02-27
TMZ provide doku for Decompiler REST API

t-m-z:
provide basic infortmation about the structure of the project
my son will upload then further information and the servlet


t-m-z:
mishap regarding directory

PR #955, opened 2022-03-09
MD brainfk derivate interpreter changed to BRAINFK_DERIVATES

Mike-3:


S-Man42:
Committed my version. Maybe you could check it, if I missed something...

PR #996, opened 2022-03-28
Time to degree

Mike-3:


lib/logic/tools/science_and_technology/time_degree.dart
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    return (sign < 0 ? '-' : '') +
+        _hours.toString() + ":" +
+        _minutes.toString() + ":" +
+        _secondsStr;
+  }
+}
+
+class DEG {
```
S-Man42, line 102: is it possible to make this private? Otherwise give it more restrictive name to get not confused with the "real" global DEG coordinate class... maybe "TimeDegreeDEG" or something.

S-Man42: Do you really need this here? Why not simply using the already existing DEG class? Isn't it code duplication here?

lib/widgets/tools/science_and_technology/time_degree.dart
```diff
@@ -0,0 +1,301 @@
...
+          ? _buildDegRow()
+          : _buildHmsRow(),
+
+        GCWMultipleOutput(children: _buildOutput())
+      ],
+    );
+  }
+
+  Widget _buildHmsRow() {
+    return Column(children: [
```
S-Man42, line 113: there is already a GCWDateTime input widget. I'd prefer to reuse and maybe expand this instead of creating something completely new.

lib/widgets/registry.dart
```diff
@@ -1070,6 +1071,12 @@ initializeRegistry(BuildContext context) {
       'symbol_murraybaudot',
       'symbol_baudot'
     ]),
+    GCWTool(tool: TimeDegree(), i18nPrefix: 'time_degree', categories: [
+      ToolCategory.SCIENCE_AND_TECHNOLOGY
```
S-Man42, line 1075: Does it make sense to put this into the datetime selection?

lib/widgets/tools/science_and_technology/time_degree.dart
```diff
@@ -0,0 +1,301 @@
...
+              flex: 6,
+              child: Container(
+                child: GCWIntegerTextField(
+                    hintText: 'DD',
+                    textInputFormatter: CoordsIntegerDegreesLatTextInputFormatter(allowNegativeValues: false),
+                    controller: _decDegreesController,
+                    onChanged: (ret) {
+                      setState(() {
+                        _currentDecDegrees = ret['text'];
+                        //_setCurrentValueAndEmitOnChange();
```
S-Man42, line 214: please remove not necessary code

test/logic/tools/science_and_technology/time_degree.dart
```diff
@@ -0,0 +1,87 @@
...
+        if (_actual == null)
+          expect(_actual, elem['expectedOutput']);
+        else
+          expect(_actual.toString(), elem['expectedOutput']);
+      });
+    });
+  });
+
+  group("DEG.parse:", () {
+    List<Map<String, dynamic>> _inputsToExpected = [
```
S-Man42, line 49: please try to test more edge cases. 
0
-0
90
-90
180
-180
360
-360
720
-720

What about roundings at the times?

59.999999999 Minutes = 60 Minutes = 1 hour 0 Minutes
61 Minutes?
25 Hours, ...

lib/widgets/main_view.dart
```diff
@@ -629,6 +630,7 @@ void _initStaticToolList() {
       className(TeletypewriterSelection()),
       className(TeletypewriterPunchTape()),
       className(TimeCalculator()),
+      className(TimeDegree()),
```
S-Man42, line 633: Maybe the name is not specific enough... What time? In the description you are naming something with "equatorial time" or similar... Maybe this should be "EquatorialTimeToDegrees" or something (rename files as well, please)

Btw. is there a source for the calculation/algorithm?

lib/logic/tools/science_and_technology/time_degree.dart
```diff
@@ -0,0 +1,146 @@
...
+  double degress;
+
+  DEG(double degress) {
+    this.degress = degress;
+  }
+
+  static DEG parse(String input, {wholeString = false}) {
```
S-Man42, line 109: Does it make sense to output not only DMM, but also DEC and DMM? For me it seems so...

lib/logic/tools/science_and_technology/time_degree.dart
BimGitHub2: I actually wanted to use the DEG class. But unfortunately that didn't work. Hence the new class.
I rename them. Private does not work because of widget and test.

lib/logic/tools/science_and_technology/time_degree.dart
BimGitHub2: I just had it in there too. But LatLon only allows angles of 90 degrees.

lib/logic/tools/science_and_technology/time_degree.dart
S-Man42: Nonetheless much code duplication: The Pattern, the rounding algorithm... Please try to avoid the duplication nonetheless...

Mike-3:
refactoring datetime_picker

changed features:
time_calculator
day_calculator
shadow_length
moon_postion
moon_rise_set
sun_position
sun_rise_set
weekday

lib/widgets/common/gcw_datetime_picker.dart
```diff
@@ -1,12 +1,15 @@
...
 import 'package:gc_wizard/widgets/common/base/gcw_text.dart';
 import 'package:gc_wizard/widgets/common/gcw_double_spinner.dart';
 import 'package:gc_wizard/widgets/common/gcw_dropdown_spinner.dart';
 import 'package:gc_wizard/widgets/common/gcw_integer_spinner.dart';
+import 'package:gc_wizard/widgets/common/gcw_toolbar.dart';
 import 'package:gc_wizard/widgets/utils/common_widget_utils.dart';
+import 'package:gc_wizard/widgets/tools/coords/base/gcw_coords_sign_dropdownbutton.dart';
 
-enum DateTimePickerType { DATETIME, DATE_ONLY, TIME_ONLY }
+enum DateTimePickerConfig {SIGN, DATE, DAY, TIME, SECOND_AS_INT, TIME_MSEC, TIMEZONES }
```
S-Man42, line 12: Please add comments to understand the options

lib/widgets/registry.dart
```diff
@@ -944,6 +945,12 @@ initializeRegistry(BuildContext context) {
     ], searchKeys: [
       'reverse',
     ]),
+    GCWTool(tool: RightAscensionTimeToDegree(), i18nPrefix: 'right_ascension_time_to_degree', categories: [
+      ToolCategory.SCIENCE_AND_TECHNOLOGY
```
S-Man42, line 949: please remove the category, as it is part of the Date Selection; Move the entry down to the date section

lib/widgets/tools/science_and_technology/date_and_time/right_ascension_time_to_degree.dart
```diff
@@ -0,0 +1,251 @@
...
+  void dispose() {
+    _hoursController.dispose();
+    _minutesController.dispose();
+    _secondsController.dispose();
+    _mSecondsController.dispose();
+
+    _decDegreesController.dispose();
+    _decMilliDegreesController.dispose();
+
+    _hoursFocusNode.dispose();
```
S-Man42, line 71: If focus was never set (e.g. nothing clicked) this results in a NullPointer exception!

```
======== Exception caught by widgets library =======================================================
The following NoSuchMethodError was thrown while finalizing the widget tree:
The method 'dispose' was called on null.
Receiver: null
Tried calling: dispose()

When the exception was thrown, this was the stack: 
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:38:5)
#1      RightAscensionTimeToDegreeState.dispose (package:gc_wizard/widgets/tools/science_and_technology/date_and_time/right_ascension_time_to_degree.dart:71:21)
#2      StatefulElement.unmount (package:flutter/src/widgets/framework.dart:4983:11)
====================================================================================================
```

lib/widgets/tools/science_and_technology/date_and_time/right_ascension_time_to_degree.dart
```diff
@@ -0,0 +1,251 @@
...
+
+      var _rightAscension = RightAscension(sign, hours, minutes, _secondsD);
+
+      var entry = <String>[i18n(context, 'common_unit_angle_deg_name'), raTime2Degree(_rightAscension).toString()];
+      output.add(entry);
+    }
+    return columnedMultiLineOutput(context, output, flexValues: [1, 1]);
+  }
+
+  _parse(String input) {
```
S-Man42, line 224: Please return a message if entry was not parsed. This supports the user, who currently might think of clicked not correctly.

lib/utils/common_utils.dart
```diff
@@ -35,6 +35,16 @@ int extractIntegerFromText(String text) {
   return int.tryParse(digits);
 }
 
+int separateDecimalPlaces (double value) {
```
S-Man42, line 38: Please always add tests for common utils!

lib/utils/common_utils.dart
```diff
@@ -182,6 +194,22 @@ String formatHoursToHHmmss(double hours) {
   return '$hourStr:$minutesStr:$secondsStr';
 }
 
+String formatDurationToHHmmss(Duration duration, {days: true, milliseconds: true, limitHours = true}) {
```
S-Man42, line 197: Tests for common utils.

lib/utils/common_utils.dart
```diff
@@ -156,18 +166,20 @@ DateTime hoursToHHmmss(double hours) {
   return DateTime(0, 1, 1, h, min, sec, milliSec);
 }
 
-String formatHoursToHHmmss(double hours) {
+String formatHoursToHHmmss(double hours, {milliseconds: true, limitHours = true}) {
```
S-Man42, line 169: Tests for common utils

lib/widgets/utils/file_utils.dart
```diff
@@ -569,8 +569,8 @@ Future<Uint8List> createZipFile(String fileName, String extension, List<Uint8Lis
 
     encoder.close();
 
-    var bytes = File(encoder.zip_path).readAsBytesSync();
-    await File(encoder.zip_path).delete();
+    var bytes = File(encoder.zipPath).readAsBytesSync();
```
S-Man42, line 572: Is this still correct?

Mike-3: line was not deleted ???

Mike-3:
all comments fixed

S-Man42:
Resolved conflict.

S-Man42:
For widget: Why don't allow all coordinate types? So, if somebody uses "normal" DMM format, he needs to manually convert it before.

I would take the normal GCWCoords widget here an maybe write the Degree in the output. Then your columnedMultilineOutput makes sense ;)

Same with Decryption: Allow to choose the output format. Then print the DEG and the chosen one.

Mike-3:
Unfortunately I don't quite understand what you mean. Nowhere is it displayed in DMM format.
grafik
grafik

S-Man42:
Powerpoint Action!!!

The idea is: For Geocachers the most common usecase would be to calculate a coordinate (mostly in DMM format finally). So let's do the step for them directly!

I think, it makes sense to allow at least (DEC, DMM, DMS) for input. Most other "real" coord formats combine both coordinate components, so they are not really useful here (Btw: Does +/- choice really make sense in context of a time? Just asking, have no idea..)

Decryption:

grafik

Encryption:

grafik

An please don't use the columnedMultiline if you just have 1 output line to stay convenient with all other tools ;)

Mike-3:
So minus makes sense, since it results in negative angles.
So I'm actually also in favor of reformatting the numbers into a coordinate format. I just didn't put it in because you said it wasn't really defined.
Which is true as far as I know.
But I still like to do the conversion.

PR #1003, opened 2022-04-04
enclosed numbers add option only numbers

Mike-3:
add split on special chars

### DIFF
--- a/assets/i18n/en.json
+++ b/assets/i18n/en.json
@@ -1898,6 +1898,7 @@
   "enclosedareas_description": "Number of holes in letters and numbers",
   "enclosedareas_with4": "Closed 4",
   "enclosedareas_without4": "Open 4",
+  "enclosedareas_only_numbers": "only Numbers",
   "enigma_title": "Enigma",
   "enigma_description": "German Military Encryption Machine in World War II.",
   "enigma_example": "SJDSF TTVWB ZXPDU",
--- a/assets/i18n/en.json
+++ b/assets/i18n/en.json
@@ -1898,7 +1898,7 @@
   "enclosedareas_description": "Number of holes in letters and numbers",
   "enclosedareas_with4": "Closed 4",
   "enclosedareas_without4": "Open 4",
-  "enclosedareas_only_numbers": "only Numbers",
+  "enclosedareas_only_numbers": "Only Numbers",
   "enigma_title": "Enigma",
   "enigma_description": "German Military Encryption Machine in World War II.",
   "enigma_example": "SJDSF TTVWB ZXPDU",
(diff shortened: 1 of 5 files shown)


lib/logic/tools/crypto_and_encodings/enclosed_areas.dart
```diff
@@ -63,24 +65,38 @@ final _AlphabetMap = {
...
 
     return alphabetMap;
-  } else
-    return _AlphabetMap;
 }
 
-String decodeEnclosedAreas(String input, bool with4) {
+String decodeEnclosedAreas(String input, bool with4, {onlyNumbers: false}) {
   if (input == null || input == '') return '';
+  var splitChars = {'+','-','/',r'\','(','),r''','{','}','[',']','<','>','!','§',r'$','%','&','=',
```
S-Man42, line 81: Why is this explicit enumeration necessary? What if another strange character occurs? Cant you just split. by any not letter/digit? What if the entire text should be counted? (and only split by spaces)?

lib/logic/tools/crypto_and_encodings/enclosed_areas.dart
```diff
@@ -63,24 +65,38 @@ final _AlphabetMap = {
...
+    } else {
+      inputSeparated = input[i] + inputSeparated;
+      lastSplitChar = true;
+    }
+  }
 
-  return input
-      .split(' ')
+  return inputSeparated
+      .split(RegExp(' '))
```
S-Man42, line 99: Split by a aingle character doesnt need a Regexp object. However: what about linebreaks or tabs? Or groups of two or three spaces? Best split is always: RegExp(r'\s+').

assets/i18n/en.json
```diff
@@ -1898,6 +1898,7 @@
   "enclosedareas_description": "Number of holes in letters and numbers",
   "enclosedareas_with4": "Closed 4",
   "enclosedareas_without4": "Open 4",
+  "enclosedareas_only_numbers": "only Numbers",
```
S-Man42, line 1901: Please start with upper case

lib/logic/tools/crypto_and_encodings/enclosed_areas.dart
Mike-3: If I only separate at spaces, the desired solution from AndyPuma's example does not come out.
Also, I don't have a bar with all the letters [A-Z] is not enough, e.g. because of (ë, ...).

lib/logic/tools/crypto_and_encodings/enclosed_areas.dart
Mike-3: Since I only explicitly insert a ' '. I actually only need to separate this one.

PR #1004, opened 2022-04-04
Substitution Breaker add dutch

Mike-3:


lib/widgets/tools/symbol_tables/symbol_replacer/symbol_replacer.dart
```diff
@@ -408,6 +408,7 @@ class SymbolReplacerState extends State<SymbolReplacer> {
       _breakerAlphabetItems = {
```
S-Man42, line 408: doesnt it make sense to reuse the original BreakerAlphabetItems instead of duplication?

S-Man42:
I'd like to put it into the 2.2.1 patch release (it's somehow a bug, isn't it? :P) Please create a PR onto 2.2.1 branch! :)

Mike-3:
You can do that in 2.2.1, but it's not a bug fix. This extension was neither forgotten nor was it buggy.

PR #1035, opened 2022-04-20
Issue Convert from/into Quadtree/NAC not working in web

Mike-3:
no problems with NAC found


lib/logic/tools/coords/converter/quadtree.dart
```diff
@@ -13,7 +13,7 @@ Quadtree latLonToQuadtree(LatLng coord, {int precision: _DEFAULT_PRECISION}) {
   var siny = sin(degreesToRadian(coord.latitude));
   var y = (_TILESIZE / 2.0) + 0.5 * log((1.0 + siny) / (1.0 - siny)) * -(_TILESIZE / (2.0 * pi));
 
-  var countTiles = 1 << precision;
+  var countTiles = int.parse('1' + '0' * precision, radix: 2); //1 << precision (bugfix for web version)
```
S-Man42, line 16: please enhance the comment with: 

// Original code: ... (must be changed due to web version issues)

PR #1036, opened 2022-04-21
tmz 2.3.0 bugfix chef

t-m-z:
https://www.geocaching.com/geocache/GC80J38_aak13-18-weihnachtsstrudel-lecker-lecker
uses ü as well as ue so umlaute have to normalized


lib/logic/tools/crypto_and_encodings/esoteric_programming_languages/chef_language/chef_language.dart
```diff
@@ -320,16 +321,18 @@ bool isValid(String input) {
 List<String> interpretChef(String language, recipe, input) {
   if (recipe == null || recipe == '') return new List<String>();
 
-  return decodeChef(language, recipe, input);
+  return decodeChef(language, normalizeUmlauts(recipe.toLowerCase().replaceAll('  ', ' ')), input);
```
S-Man42, line 324: What happens with three spaces or a space and enter or similar? The best why to handle this is:

`replaceAll(RegExp(r'\s+'), ' ')`

Or at least (if not all whitespace should be affected):

`replaceAll(RegExp(r' +'), ' ')`

(but then think about Tabs, etc...)

S-Man42: Still open, I guess.

t-m-z: Multiple spaces are shrinked to one space.
Tabs should not be a problem.
CRLF is handeled seperately.

Attention: the recipe is faulty! 

- The method "Dann alles kräftig durchrühren." does not exist. The method has to be deleted.
- The method "... schmelzen" is missing. Hence the recipe produces "78 32 52 56 0 32 50 0 46 49 48 50 39 32 0 32 49 48 176 32 53 50 46 57 49 54 39" which obviously means "N 48 2.102' 10° 52.916'"

PR #1039, opened 2022-04-24
tmz 2.2.2 bugfix numeral words

t-m-z:
substitute linefeed with space - otherwise linefedd will be recognized as character and causes trouble


lib/logic/tools/crypto_and_encodings/numeral_words.dart
```diff
@@ -2315,6 +2315,7 @@ List<NumeralWordsDecodeOutput> decodeNumeralwords(
 
     // trim korean
     input = input
+        .replaceAll('\n', ' ')
```
S-Man42, line 2318: Best way to handle white space is...

replaceAll(RegExp(r'\s+',' '))

Does it make sense here?

t-m-z: sounds reasonable

PR #1042, opened 2022-04-26
Esoteric language piet

Mike-3:
no better solution for generator found


Mike-3:
"Gemälde Piet Mondrians"  is not a piet prgram.
Piet_Program.gif does not work on my reference website either (https://www.bertnase.de/npiet/npiet-execute.php)

Mike-3:
Piet_Program.gif also does not work with https://gabriellesc.github.io/piet/  (infinite loop)

PR #1044, opened 2022-05-03
tmz 2.2.2 bugfix calendar

t-m-z:
catch null pointer


S-Man42:
You changed the logic. Is it possible to add some tests?

PR #1052, opened 2022-05-10
Added Reversewhereigo at searchstrings

Birdy2404:


S-Man42:
a) en.json is not the correct point for it. In that case, it is common.json ;) BC you use it in German as well, it's not a pure English term
b) in Common following terms are already existing: reversewherigo reversewhereigo; so the original issue to not find the tool with "reverse" can either be a user mistake or a bug at another point. For me it works without any problems. One can think about adding "reversewig" to the list nonetheless.
c) Since it is a bugfix, pleass create a PR onto the bugfix branch 2.2.2, nit onto 2.3.0

😉

PR #1054, opened 2022-06-13
music notes segmentdisplay

Mike-3:


S-Man42:
Cannot compile this:
```

lib/widgets/tools/science_and_technology/segment_display/base/n_segment_display.dart:72:19: Error: No named parameter with the name 'gesturesToOverride'.
                  gesturesToOverride: [GestureType.onTapDown],
                  ^^^^^^^^^^^^^^^^^^
/C:/Users/s-man/AppData/Local/Pub/Cache/hosted/pub.dartlang.org/touchable-1.0.0/lib/src/canvas_touch_detector.dart:13:9: Context: Found this candidate, but the arguments don't match.
  const CanvasTouchDetector({Key? key, required this.builder}) : super(key: key);
        ^^^^^^^^^^^^^^^^^^^
```

Mike-3:
Merged with 2.3.
No problems found. CanvasTouchDetector works.

S-Man42:
Does "Encryption" makes really sense here? Because it always prints the first occurence (all notes exist 2 or 3 times), which is mainly too "low" (meaning outside the main lines). So, it will never create a "real" music line... I believe it is not very helpful. What do you mean?

Mike-3:
I agree with you, but I would find an encryption practical. we have it in almost everywhere else.
One could try to use the notes on the main lines and not just the first match.

S-Man42:
yes. That would be the solution. Always show the ones which occur within the main lines. That would be ok for me

PR #1055, opened 2022-06-14
Bugfix Issue: Hidden Data: When Archive in Archive, File are showed more than once bug 

Mike-3:


S-Man42:
Looks great at the first glance! Very well done!

However, now I am missing the correct Source paths:

grafik

As decribed here: https://user-images.githubusercontent.com/16626063/151339058-8e41e8d8-b941-49c6-9b8e-b5d8804699ff.png

The path should be:

Homer: BZIP2 -> Bzip.tar
Donald / Garfield: BZIP2 -> Bzip.tar -> Temp.zip 

This is for knowing about the archive structure. I guess, it's possible to fix, isn't it?

Mike-3:
Sorry, but I haven't found a solution to also display Bzip.tar, it only worked when I unpack the gz and tar stream in a loop. I tried it out for a long time.

S-Man42:
Something's still not correct here:

grafik

1. Why are there two same entries (1 and 2 in the image)?
2. Why "Temp.tar.tar" instead of "Temp.tar"?
3. "Quelle:" is only for the beginning, not for inside the path: "Quelle: thepath"
4. So: I'd expect: "Quelle: BZIP2 -> Temp.tar -> Temp.zip" because the image "donald.jpg" is inside the Temp.zip file which is inside the Temp.tar file. No more. The simple "ZIP" or "TAR" should only be fallbacks, if there's no filename.

S-Man42:
I checked the input file structure:

Actual:

```
[
  {
    path:null; name:<<!!!HIDDEN_FILE!!!>>_0; bytes:107211; children:[
      {
        path:null; name:Temp.tar.tar; bytes:118784; children:[
          {
            path:null; name:<<!!!HIDDEN_FILE!!!>>_0; bytes:117248; children:[
              {
                path:null; name:homerl.jpg; bytes:71128; children:[
                  
                ]
              },
              {
                path:null; name:Temp.zip; bytes:45060; children:[
                  {
                    path:null; name:<<!!!HIDDEN_FILE!!!>>_0; bytes:45056; children:[
                      {
                        path:null; name:donald.jpg; bytes:17708; children:[
                          
                        ]
                      },
                      {
                        path:null; name:garfield.jpg; bytes:27696; children:[
                          
                        ]
                      }
                    ]
                  }
                ]
              }
            ]
          }
        ]
      }
    ]
  }
]
```

Expected:

```
[
  {
    path:null; name:Temp.tar.bz2; bytes:107211; children:[
      {
        path:null; name:Temp.tar; bytes:118784; children:[
          {            
            {
              path:null; name:homerl.jpg; bytes:71128; children:[
                  
              ]
            },
            {
              path:null; name:Temp.zip; bytes:45060; children:[                
                {
                  path:null; name:donald.jpg; bytes:17708; children:[
                        
                  ]
                },
                {
                  path:null; name:garfield.jpg; bytes:27696; children:[
                          
                  ]
                }                 
              ]
            }
          }
        ]
      }
    ]
  }
]
```

1. Why is first level a "Hidden File"? It is clearly not, because it's the given file. It has the name "Temp.tar.bz2"
2. Why does the second level has the name "Temp.tar.tar" instead of "Temp.tar"?
3. Why does Temp.tar have a "Hidden File" level. It has not. It simply contains "homer1.jpg" and "Temp.zip" without an intermediate structure.
4. Same as (3) with "Temp.zip": There is no intermediate level. Just two JPG files.

I guess, when finding the problems here, all issues pointed out above are solved automatically. I believe in the logic, you put an archive automatically into a hidden file structure. But it is not automatically.

PR #1061, opened 2022-07-20
Reflection : test for new folder structure

Mike-3:
created with script
imports corrected


S-Man42:
close due to new try :)

PR #1063, opened 2022-08-10
Magic eye solver

Mike-3:


S-Man42:
1. Please resolve conflicts
2. You have an issue with the first "open file" widget. When you load an image in the "encryption" page, and then go to "decryption" you can see, that the file is still loaded, so it is the same file dialog.
3. I am not able to encrypt the files that way, that the decrypter resolves something. Maybe I have bad input data... I took:
- Texture: https://s3.amazonaws.com/easystereogrambuilder-prod/data/gallery/2/126/t.jpg
- Image: https://s3.amazonaws.com/easystereogrambuilder-prod/data/gallery/1/8600/t.jpg
(and several other data but it never worked... maybe too small images or something?)

Mike-3:
conflics solved
Hidden image was to small. Not invers hidden image.
Image is internally inverted and converted to grayscale.

Mike-3:
What else could I do that the picture is scaled up if it's too small.

S-Man42:
What about simply an error message?

Mike-3:
ok

PR #1068, opened 2022-08-24
tmz 2.3.0 befunge

t-m-z:
new feature: Befunge Interpreter and Generator


t-m-z:
checked Textfield within different tools - no bugs detected

Reason for this field "style": I needed a textcontroller with MonoSpace-Style to write Befunge Programs

PR #1070, opened 2022-08-28
tmz 2.3.0 wet bulb temperature - another index for apparent temperatures

t-m-z:
the feature provides to indices
- wet bulb temperature according to https://www.omnicalculator.com/physics/wet-bulb
- wet bulb globe temperature, calculated with an approximate formula according to https://www.weather.gov/media/tsa/pdf/WBGTpaper2.pdf

Regarding the WBGT I wait for some answers to confirm my assumptions. However the formula is correct and the compiled table is similar to the one I found in the internet at  https://climate-preparedness.com/understanding-wet-bulb-temperature-and-why-it-is-so-dangerous/ and https://ksi.uconn.edu/prevention/wet-bulb-globe-temperature-monitoring/


S-Man42:
Means: Not merging and still waiting?

t-m-z:
I suggest check for code smell and merge. Because there are such a lot of different interpretations and formulas for the WBGT the results of the feature are ok.

t-m-z:
In parallel, I will make a proposal for improvement for the Bundeswehr.
A friend of mine is a department head at the Software Competence Center and responsible for software development. Let's see if I can get another look at the mathematical basics this way.

assets/i18n/en.json
```diff
@@ -1637,6 +1672,27 @@
   "coords_formatconverter_naturalareacode_y": "Y (North)",
   "coords_formatconverter_openlocationcode_novalidcode": "No valid OpenLocationCode input",
   "coords_formatconverter_slippymap_zoom": "Zoom",
+  "coords_formatconverter_w3w_language" : "Language",
```
S-Man42, line 1675: you inserted w3w here

lib/logic/tools/science_and_technology/apparent_temperature/wet_bulb_temperature.dart
```diff
@@ -0,0 +1,63 @@
...
+  if (temperatureUnit == TEMPERATURE_FAHRENHEIT) temperature = (temperature - 32) / 1.8;
+
+  // https://rechneronline.de/air/wet-bulb-temperature.php
+  double WBT = temperature * atan(0.151977 * sqrt((humidity + 8.313659))) +
+      atan(temperature + humidity) -
+      atan(humidity - 1.676331) +
+      0.00391838 * pow(humidity, 1.5) * atan(0.023101 * humidity) -
+      4.686035;
+
+  if (temperatureUnit == TEMPERATURE_FAHRENHEIT) {
```
S-Man42, line 58: Please dont calculate different units in one method. Moreover if you want to display different units, use the already existing internal unit converter in the frontend for this

assets/i18n/en.json
t-m-z: ups - don't know how this could happen.
However - I am working on that  ;-)

lib/logic/tools/science_and_technology/apparent_temperature/wet_bulb_temperature.dart
t-m-z: done - however the Widget is buggy

S-Man42:
grafik

<20% was ok, now NPE

S-Man42:
Do you really think, that "prefixes" concerning temperatures make really practical sense? Did you ever hear of MilliCelsius or MegaKelvin? I would leave it to achieve more simplicity.

t-m-z:
you are totally right, I have to tweak the widget to suppress the prefix

Sent from MailDroid

-----Original Message-----

lib/widgets/tools/science_and_technology/apparent_temperature/wet_bulb_temperature.dart
```diff
@@ -0,0 +1,167 @@
...
+          min: 0.0,
+          unitList: humidity,
+          onChanged: (value) {
+            setState(() {
+              _currentHumidity = value;
+            });
+          },
+        ),
+        GCWTextDivider(text: i18n(context, 'common_outputunit')),
+        GCWUnits(
```
S-Man42, line 55: I believe instead of GCWUnits, you should use GCWUnitDropDownButton. This doesn't include the prefixes

t-m-z:
hm. thanks. this leads me to a point I stood last winter: writing an API for GCW

I lost grip on all the Widgets, methods, constants, etc.

Sent from MailDroid

-----Original Message-----

t-m-z:
Thanks for the hint with the UnitDropDown
I still got this exception - but I think I found the bug in suppressing the overflow

S-Man42:
the UnitDropDown has the option: "onlyShowSymbols"

You could set it to true, which looks better at this width :)

Did you check how you colored symbols look with the bright theme? i can imagine, that for instance the light blue is not clearly visible. Maybe one need to find something to increase the contrasts...

S-Man42:
Currently you have no opportunity to copy your result (beside manually marking it, of course).
Maybe it makes sense to think about two GCWDefaultOutputs: One for the pure number: "23.15 °C; copyText: "23.15112312 <whatever the calculation really returns unformatted>" and one for the hint

t-m-z:
Thank you for those hints.
Instead of GCWDefaultOutput I used GCWOutput and rearranged the colums. This should fit better.
The colors are OK - however I added a backgroundcolor depending on the theme to increase tho contrast.
It's true, onlyShowSymbols should be false which looks better at this width :-)

PR #1071, opened 2022-08-28
add symbols

t-m-z:
Checked all and added some - however this is a tricky topic which allows a lot of interpreting.
The source from the internet ard all different.
I tried to get the greatest common interpretation


S-Man42:
only a minor note: If you have 2 A's or something, you don't need to use the "special_mappings". Just name it "A_". The logic will transform it into a second A internally.

t-m-z:
understood

Issue #828, opened 2021-12-22
Combine "Piano Key Frequences" with the Symboltable of Notenames to generate sound


Birdy2404:
For guys like me, who have a bis issue with music it would be great to combine the two informations. As we have the frequences and could type in the notes by the existing symbol table to solve mysteries we "just need" the possibility to create sound and play it to find the required song.

S-Man42:
![grafik](https://user-images.githubusercontent.com/16626063/188335882-b09d6663-9e57-4b5e-9f39-b85c83763169.png)
![grafik](https://user-images.githubusercontent.com/16626063/188335890-ff5d3a60-23ac-48a2-a57f-216a86734e1f.png)

S-Man42:
so much ideas...

adding a "metronome" slider to change play speed. (variable base tone length)

Play sound with GCWSoundPlayer would add timer and position slider

Highlight the currently playing note...

Birdy2404:
Calculating the length accorfing bpm:
https://rechneronline.de/musik/notenlaenge.php

BPM-areas for italian speed names (allegro, ...) 
https://www.musicca.com/de/fachbegriffe

S-Man42:
Grmlgrmlgrmlgrml

S-Man42:
Hm, your table is very inexact. What should the tool return on 74 BPM? adagio or adagietto?

Birdy2404:
Wikipedia is even worse. Depending on the metronome type the speed is changing 🙈
I have the feeling, we need an expert for this.

S-Man42:
I have the feeling, we shouldnt make it too complicated. In the end, it should be a tool for entering some simple notes given in a music mystery and hope you'll hear something. it's not a professional music maker...

S-Man42:
It seems, the most problematic step is creating a sound by flutter. I cannot find a sound generator library (only one old and only suitable for Android)

PR #1073, opened 2022-09-02
Update notes segmentdisplay

Mike-3:


S-Man42:
Is it possible to create an output without gaps here? And maybe thinner symbols (without that much empty line space around).

So that would create something which looks more like a real note sheet, not just some lonely note symbols. I guess that would be a great preparation for a potential output of this one here: #828

S-Man42:
Update: Part 1: No symbol spaces:

grafik

Todo: Thinner segment display output symbols...

S-Man42:
Tadaaa...

grafik

S-Man42:
Now missing for entire perfection: The chosen note clef (violin, alt, bass) always at the beginning.

Mike-3:
Very nice. Have you a example with b or #.

S-Man42:
Have look at the last both. or what do you mean

PR #1075, opened 2022-09-03
Feature day of the year

Mike-3:


S-Man42:
Merged too early... 

Got a mess of Exceptions from the DatePicker widget:

grafik

Cannot say exactly when it happens... Just switch around, play with buttons and focus. I guess some TextEditingControllers are not initiated correctly within the different options...

S-Man42:
Furthermore: I have extreme issues with day entering... cannot use backspace on the day field, etc. Maybe it's not a good idea to integrate this into the GCWDatePicker. I would leave it outside. It will never be used in another tool I guess and it makes the GCWDatePicker more complicated than necessary...

PR #1076, opened 2022-09-03
tmz 2.3.0 bundeswehr authentification

t-m-z:
provides a toolset which utilzes the Bundeswehr Sprechtafle to authenticate and obfuscate text as described in
 https://www.reservistenverband.de/wp-content/uploads/2019/08/taschenkarte_fernmeldedienst_aller_truppen_nr.9_3.pdf
 https://www.rk-ubstadt.de/?p=2079
 https://www.geocaching.com/geocache/GC7DF7R_garnisonsgeschichte-6?guid=27ba9b97-31ab-473e-b35f-6d657e5d60b4
 https://fschjgbtl251.de/technik/fernmeldetechnik/
 http://scherfel.de/shared-files/388/FmDst-SchleiernUndSchluesseln.pdf


assets/i18n/en.json
```diff
@@ -415,6 +415,43 @@
   "brainfk_input_close": "] Jump back to matching [",
   "brainfk_error_pointertoosmall": "Pointer too small",
   "brainfk_error_customundefined": "Custom not defined",
+  "bundeswehr_title": "Bundeswehr Talking board",
```
S-Man42, line 418: I can image "Bundeswehr" is much more than just "Sprechtafel". So please make the identifier more unique.

assets/searchstrings/common.json
```diff
@@ -62,6 +62,9 @@
   "bookcipher": "",
   "braille": "braille",
   "braille_euro": "eurobraille computerbraille iso8859-1 din32982 ansi",
+  "bundeswehr": "zdv 55/512 ar-950/0-0-18",
```
S-Man42, line 65: special characters not necessary anymore

lib/logic/tools/crypto_and_encodings/bundeswehr_talkingboard/bundeswehr_auth.dart
```diff
@@ -0,0 +1,249 @@
...
+// https://www.rk-ubstadt.de/?p=2079
+// https://www.geocaching.com/geocache/GC7DF7R_garnisonsgeschichte-6?guid=27ba9b97-31ab-473e-b35f-6d657e5d60b4
+// https://fschjgbtl251.de/technik/fernmeldetechnik/
+// http://scherfel.de/shared-files/388/FmDst-SchleiernUndSchluesseln.pdf
+//
+// 08 41 75 51 36 30 99 28 65 10 66 46 55 20 79 16 82 02 93 45 97 19 88 81 63 01 38 72 29 87 98 54 50 04 49 44 27 31 77 17 18 64 15 56 32 86 73 26 67 95 35 14 96 39 06 80 25 68 85 43 12 48 07 24 74
+// cjosmhlquweby
+// dainrtxfkpvzg
+
+class AuthentificationTable {
```
S-Man42, line 12: Since it is public, please add identifier before class names and other members. I guess "AuthTable" could. be wever, where else.

t-m-z:
I hope the changes made it

S-Man42:
You are sure about this feature in general? It is nothing that is sensitive or could take you into "devil's kitchen" at all?

S-Man42:
Authentication: First click -> Exception (tried to increase the "letter")
grafik

Since I have no idea of the code, but mostly that should be avoided: Is it allowed to enter the same letter several times? 
grafik


The Obfuscation:

Encryption Input was: dsfsdvxcvsdsdfsdfsd
Encryption Output was: FH BL OE ZK VM JC LV HM KD HK MR QN DK JC LV OE LV MV JC LV MV JC LV
->
put this into decryption Input: 
Decryption output was: 3323131133131231231

Is that correct?

When I enter the crypto text without spaces, the output cannot be calculated:
grafik

This should be done more user friendly. Best way is, to insert the spaces manually internally (theres a function in custom utils for it). Moreover, you can additionally prepare the textfield to format the input for you automatically if you like (have a look at Enigma's plugboard textfield for this)

S-Man42:
Additional thought to the "Random" thing. As said, I dont really know anything about this code. However: Does an "Random" DEcryption field make sense? As I see it, the decryption always needs a proper key in order to do the decryption properly. Why should someone to the decryption with a random key which never fits the encryption key...

lib/logic/tools/crypto_and_encodings/bundeswehr_talkingboard/bundeswehr_auth.dart
```diff
@@ -0,0 +1,249 @@
...
+  'V',
+  'W',
+  'X',
+  'Y',
+  'Z'
+];
+
+const DIGITS = ['1', '2', '3', '4', '5', '6', '7', '8', '9', '0'];
+
+const AUTH_RESPONSE_OK = 'bundeswehr_talkingboard_auth_response_ok';
```
S-Man42, line 87: Please ensure prefixes or private members here as well. The problem is, that otherwise, the auto complete function of the IDEs show you this member if someone tries to write AUTH_XYZ, which is quite imaginable. Unfortunately, Dart has no package private concept, so things are only private to the own file or public to everything. So it must be ensured, that there is no confusion in any way. So, members with names, that could be part of anything, shall be named with a unique prefix, to avoid taking wrong members from different functions

S-Man42:
You did not fix this. It is still there:

> Authentication: First click -> Exception (tried to increase the "letter") grafik
> 
> Since I have no idea of the code, but mostly that should be avoided: Is it allowed to enter the same letter several times? grafik
> 
> The Obfuscation:
> 
> Encryption Input was: dsfsdvxcvsdsdfsdfsd Encryption Output was: FH BL OE ZK VM JC LV HM KD HK MR QN DK JC LV OE LV MV JC LV MV JC LV -> put this into decryption Input: Decryption output was: 3323131133131231231
> 
> Is that correct?
> 
> When I enter the crypto text without spaces, the output cannot be calculated: grafik
> 
> This should be done more user friendly. Best way is, to insert the spaces manually internally (theres a function in custom utils for it). Moreover, you can additionally prepare the textfield to format the input for you automatically if you like (have a look at Enigma's plugboard textfield for this)

S-Man42:
I would take a normal GCWTextDivider instead of GCWExpandableTextDivider here:

grafik

It looks less "gequetscht" und since it is a necessary input, i guess, there's no reason for collapse/hide these fields.

S-Man42:
Per default your output is "ERROR". That should not be. If nothing entered, we always show no output. Please stay convenient with the rest of the app. "ERROR" might confuse the user. Maybe it is worth thinking about not calling it "ERROR" that explicitly. Maybe something like a hint: "Letters missing" or "Letters duplicated" or whatever. It's not like a bug in the code, it should be more like guidance. Otherwise the user will get the "ERROR" message crying at his face although he is currently typing. That might unsure him.

PR #1079, opened 2022-09-08
Typo Music Notes --> Musical Notes acc Wikipedia

Birdy2404:
https://en.wikipedia.org/wiki/Musical_note


Birdy2404:
Maybe also change the music_notes_title --> musical_notes_title ?

PR #1086, opened 2022-09-10
add calculations for irrational numbers sqrt 2, 3, 5

t-m-z:
if there is some time - otherwise postpone to 2.4.0


assets/searchstrings/de.json
```diff
@@ -338,6 +338,7 @@
   "silverratiosearch": "",
   "skytale": "skytale stock stab",
   "solitaire": "kartenspiel kartendeck stromchiffrierung solitaer",
+  "sqrt": "quadratwurzel wurzel",
```
S-Man42, line 341: Just minor: "wurzel" is contained in "quadratwurzel" and therefore not necessary.

t-m-z: it is and will be a mystery

PR #1092, opened 2022-09-14
Hidden data refactoring

Mike-3:


S-Man42:
s-man42.de/robot2.jpg

grafik

Detects a ZIP file, but no extracting here

PR #1094, opened 2022-09-15
Feature Rabbit code

Mike-3:
Input like RC4


S-Man42:
The "please enter key" is contraproductive here. It is displayed at each input character. So when I am first entering my crypto text, ever letter I enter, the toast arises. Maybe better way is either no doing anything or returning the hint as output.

Binary output: Since it is fixed number output, i guess, maybe it makes sense to leftPad the blocks always to 8 characters. So 11111 will be returned as 00011111?

S-Man42:
Is encryption and decryption done with the same tool?

assets/searchstrings/en.json
```diff
@@ -306,6 +306,7 @@
   "projectiles": "bullets projectiles ballistics ",
   "qrcode": "qr barcode",
   "quadraticequation": "quadratic equation",
+  "rabbit": "rabbit stream cipher",
```
S-Man42, line 309: secretrabbitcode secretrabbitstreamcipher

PR #1105, opened 2022-10-10
formula solver: special spaces

Mike-3:


S-Man42:
Isnt it better to name it: normalizeSpaces?

Mike-3:
I had it first, but then renamed it if other characters were added.
But of course we can also rename specifically only for spaces.

S-Man42:
I think as long as there a no others...

PR #1109, opened 2022-10-14
tmz bugfix wig hash obfuscation clitches

t-m-z:
- get also numeric hash breaks and add a second line
- identify obfuscated texts with line breaks and de-obfuscate these texts


lib/logic/tools/crypto_and_encodings/wherigo_urwigo/krevo.wherigotools/ucommons.dart
```diff
@@ -55,3 +55,16 @@ String findHash(int hashToFind, int len) {
 
   return null;
 }
+
+String findHashNumeric(int hashToFind, int len) {
```
S-Man42, line 59: I personally don't like it to change the original code from external source significantly (meaning: Not more than really necessary to get it running). The idea behind this: When there will be a change in the external sources, I try to keep it the same to integrate further changes better and find the code lines by pure text search. So, in that case: I will not be able anymore to find the "findHash" part that easy anymore. 

Adding a new method by our own should be done in our own code, not in the external one. Maybe it's a better choice to write a wrapper maybe in "urwigo_tools":

findHashAlphaNumeric() {
   return external.findHash();
}

and then new code:

findHashNumeric()

Btw: Why "alphaNumeric"? Isnt it better called "letters"? or "alphabetical"?

android/build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.5.10'
+    ext.kotlin_version = '1.7.20'
```
S-Man42, line 2: please do not check in this change. I hope this will not be necessary in my code. System changes should only be done in the base unless it is absolutly necessary for running this specific code, but I cannot see why this should be.

t-m-z: How do I checkout this file?

lib/logic/tools/crypto_and_encodings/wherigo_urwigo/krevo.wherigotools/ucommons.dart
t-m-z: check - I hope it is okay to make convBase public.

android/build.gradle
t-m-z: Okay - I changed again to the old values

t-m-z: Now I receive 
e: Incompatible classes were found in dependencies. ...
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':audioplayers_android:compileDebugKotlin'.


┌─ Flutter Fix ────────────────────────────────────────────────────────────────┐
│ [!] Your project requires a newer version of the Kotlin Gradle plugin.                                                                             │
│ Find the latest version on https://kotlinlang.org/docs/gradle.html#plugin-and-versions, then                                     │
│ update C:\Programmierung\android\build.gradle:                                                                                                          │
│ ext.kotlin_version = '<latest-version>'                                                                                                                             │
└────────────────────────────────────────────────────────────────────────┘

t-m-z: No clue about that

S-Man42: i take care for it. leave it there

PR #1110, opened 2022-10-15
tmz 2.4.0 bowling points

t-m-z:
;-)


lib/logic/tools/games/bowling.dart
```diff
@@ -0,0 +1,96 @@
+class BowlingRound {
+  final int one;
+  final int two;
+  final int three;
+
+  BowlingRound({this.one = 0, this.two = 0, this.three = 0});
+}
+
+Map<int, Map<int, int>> BOWLING_THROWS = {
```
S-Man42, line 9: This map makes no sense to me. As I see it, you need only lists from 0 to n. and not even fixed. You take n and can generate an integer list directly out of it...

S-Man42:
Please put the logic into a logic class. Calculation of the score board should be independent from the view.

Here you have a classical case for model view concept. You have 1 model logic, but two views on it (table and score board). Please build it as this.

Dont forget to add test cases to the logic

t-m-z:
Logic and Widget is seperated.
Test cases will follow.

lib/logic/tools/games/bowling.dart
t-m-z: As I see it, for the DropDown I need a map or at leat an idea how to build the necessary map from a list.

t-m-z:
I still have to test  as soon as this compiler error is fixed

PR #1115, opened 2022-10-15
tmz bugfix numeral words

t-m-z:
from the discussion at geoclub


capoaira:
Is it already in the internal Beta version?
I just updated the app, but I get a gray screen as soon as I type the first letter,

S-Man42:
Yes there is a bug in internal version. I will push a new one today

PR #1126, opened 2022-10-18
tmz refactor numeral words:

t-m-z:
- add feature to identify languages
- add default language for textserach according to locale
- add three languages: GRC, SVK, CZE and also greek/latin letters
- add old greek, old greek latin


S-Man42:
Overview -> Greek old

grafik

S-Man42:
Text Search: Trying to choose another language than EN does not work. Still EN afterwards.

PR #1130, opened 2022-10-23
enrich multiple choice answers with hases for upper and lower case

t-m-z:
A point made by AndyPuma:
when asked for Multiple Choice answers you have to push a button. Hence a broken Hash does not help.
For example  https://coord.info/GC2KHC4 evaluates all answers at the end. This means you cannot try several times.


S-Man42:
grafik

I am not quite sure... Was this issue not about showing the correct answer in MC questions? I believe I didn't get this right, but in that case (edersee) I cannot see which one is correct without reading the answer's text... Is this ok that way or not?

t-m-z:
that is the way - regarding the source codes two answers will be handled.
So the correct answers can only be found reading further information 

Sent from MailDroid

-----Original Message-----

PR #1140, opened 2022-10-28
added returns

Birdy2404:


S-Man42:
Please dont do this manually. It's the work of the internal formatter (which indeed has very strange decisions sometimes..). It ensures that the code is same formatted for every coder....

PR #1154, opened 2022-11-05
update gcw_coords_reversewherigo_day1976.dart

t-m-z:
convert input to lower case

### DIFF
--- a/lib/widgets/tools/coords/base/gcw_coords_reversewherigo_day1976.dart
+++ b/lib/widgets/tools/coords/base/gcw_coords_reversewherigo_day1976.dart
@@ -81,6 +81,6 @@ class GCWCoordsReverseWherigoDay1976State extends State<GCWCoordsReverseWherigoD
 
   _setCurrentValueAndEmitOnChange() {
     widget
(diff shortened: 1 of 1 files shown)


lib/widgets/tools/coords/base/gcw_coords_reversewherigo_day1976.dart
```diff
@@ -81,6 +81,6 @@ class GCWCoordsReverseWherigoDay1976State extends State<GCWCoordsReverseWherigoD
 
   _setCurrentValueAndEmitOnChange() {
     widget
-        .onChanged(ReverseWherigoDay1976.parse(_currentA.toString() + '\n' + _currentB.toString()));
+        .onChanged(ReverseWherigoDay1976.parse(_currentA.toString().toLowerCase() + '\n' + _currentB.toString().toLowerCase()));
   }
```
S-Man42, line 85: Wrong place. It should be done in the parse function or in the ..ToLatLon functions to avoid doing this on every call...

t-m-z:
okay - I do not really understand the difference . but thanks for the hint

S-Man42:
You have at least 3 point where you call the parse method. You can either make this toLowerCase on all those (and possible future) calls or you make it centralised at one point...

PR #1173, opened 2022-11-13
add default padding for every call of GCWSegmentDisplay as a workaround

t-m-z:
horizontalSymbolPadding: 5,
        verticalSymbolPadding: 5,

### DIFF
--- a/lib/widgets/tools/crypto_and_encodings/babylon_numbers.dart
+++ b/lib/widgets/tools/crypto_and_encodings/babylon_numbers.dart
@@ -127,6 +127,8 @@ class BabylonNumbersState extends State<BabylonNumbers> {
         segmentFunction: (displayedSegments, readOnly) {
           return BabylonNumbersSegmentDisplay(segments: displayedSegments, readOnly: readOnly);
         },
+        horizontalSymbolPadding: 5,
+        verticalSymbolPadding: 5,
         segments: segments,
         readOnly: true);
   }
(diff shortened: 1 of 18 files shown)


S-Man42:
No, to avoid multiple default values all. over the app, the right place for this would be the constructor of the GCWSegmentDisplayDingens

t-m-z:
Hm. Okay. But then I have no clue how to do this the right way. 
There is a widget.horizontalPadding which does not work properly. So I thought to set the data explicitly.


Sent from MailDroid

-----Original Message-----

S-Man42:
```
const GCWSegmentDisplayOutput(
      {Key key,
      this.upsideDownButton: false,
      this.segmentFunction,
      this.segments,
      this.readOnly,
      this.trailing,
      this.showZoomButtons: true,
      this.verticalSymbolPadding: 5,
      this.horizontalSymbolPadding: 5})
      : super(key: key);
```

This does not work?

Mike-3:
This fix works on my short test

  const GCWSegmentDisplayOutput(
      {Key key,
      this.upsideDownButton: false,
      this.segmentFunction,
      this.segments,
      this.readOnly,
      this.trailing,
      this.showZoomButtons: true,
      this.verticalSymbolPadding = 5.0,
      this.horizontalSymbolPadding = 5.0})
      : super(key: key);

  @override
  _GCWSegmentDisplayOutputState createState() => _GCWSegmentDisplayOutputState();
}

t-m-z:
unfortunately not

Sent from MailDroid

-----Original Message-----

PR #1189, opened 2022-11-29
Issue turn columned multiline function into separate widget

Mike-3:


S-Man42:
The first refactoring from you created conflicts, I guess. I would like to think about after the restructuring nonetheless. During the restructuring, I feel we need to adjust some widgets like this one either. So I'd wait for it.

PR #1194, opened 2022-12-01
Issue waldmeister input keep leading zeros

Mike-3:
Suggestion:
Solution with integer spinner


S-Man42:
I don't like the way to manipulate the input spinner for that. In most use cases you don't want to leading zeros but simply normal numbers. Please just replace the integer spinners with textfields in waldmeister... or maybe make an deviation of the spinner if you want to keep it (maybe WaldmeisterSpinner extends GCWIntegerSpinner) but I am not quite sure if this is really necessary.

PR #1201, opened 2022-12-12
bugfix web tool list scroll problem

Mike-3:


S-Man42:
The "About" page is not pannable anymore in emulator!

PR #1219, opened 2022-12-23
Json type check

Mike-3:


S-Man42:
Missing type checks for JSON Arrays and JSON Objects

lib/utils/common_utils.dart
```diff
@@ -404,3 +404,40 @@ String normalizeCharacters(String input) {
 
   return input;
 }
+
+double jsonDouble(dynamic value) {
```
S-Man42, line 408: missing tests

lib/utils/common_utils.dart
```diff
@@ -404,3 +404,40 @@ String normalizeCharacters(String input) {
...
+}
+
+bool jsonBool(dynamic value) {
+  if (value is bool)
+    return value;
+
+  return null;
+}
+
+String jsonString(dynamic value) {
```
S-Man42, line 433: I believe that's not working. Here you'd get something like Object is not Type String. I guess .toString() is necessary at least, but in my opinion a proper type check is necessary as well...

lib/utils/common_utils.dart
```diff
@@ -404,3 +404,40 @@ String normalizeCharacters(String input) {
...
+    return value;
+}
+
+List<String> jsonStringList(List<dynamic> list) {
```
S-Man42, line 437: what about int arrays and so on?

lib/utils/common_utils.dart
```diff
@@ -404,3 +404,40 @@ String normalizeCharacters(String input) {
...
+double jsonDouble(dynamic value) {
+  if (value is double)
+    return value;
+  else if (value is int)
+    return (value as int).toDouble();
+
+  return null;
+}
+
+int jsonInt(dynamic value) {
```
S-Man42, line 417: I'd see these functions in a separate json_utils

lib/utils/common_utils.dart
```diff
@@ -404,3 +404,40 @@ String normalizeCharacters(String input) {
...
+  if (value is int)
+    return value;
+  else if (value is double)
```
S-Man42, line 420: Hm... discussion here... Getting a 1.2 as input value will be handled as 1? In my opinion it's an errornous input and should be handled as this. Maybe the using tool can decide what to do in error case, but the util function never should. Or we add an jsonDoubleOrInt method separately...

Mike-3: I just really didn't want to consider inputs from 1.0 as bugs.

lib/utils/common_utils.dart
Mike-3: Wasn't necessary at the moment

lib/utils/common_utils.dart
Mike-3: ok

Mike-3:
rebase conflict error

PR #1225, opened 2022-12-24
convert string files

Mike-3:


Mike-3:
Load and save unicode string files.

PR #1235, opened 2022-12-31
Feature Number pyramid

Mike-3:


S-Man42:
1. Please follow conventions for external libs:
logic/external_libs/dennistreya/...
2. Please add GPL advice into head of external libs/ports

S-Man42:
Bei Änderung Row Count ggf. gleichen den passenden Zoom einstellen? Also, dass die gesamte untere Zeile passt oder so?

Mike-3:
Revision is done

S-Man42:
grafik
It's not possible to read anything...

"Could not be solved" is not a very good text, I find. Maybe you can say the reason to guide the user. Maybe something like "due to too many possible solutions" or whatever...

S-Man42:
I saw, you touched the PSE, Grids, Sudoku, etc... What do you think: Does it make sense to add the zoom buttons there as well? Especially Game of Life, Grid and PSE can be extremely small, so it could be a good enhancement.

S-Man42:
> It's not possible to read anything...

After thinking about this for a while, I believe we don't come around a separate input dialog: Tap on the field, open a small dialog with textfield and button to enter the numbers. Otherwise the textinput could get extremely small.

Mike-3:
> grafik It's not possible to read anything...
> 
> "Could not be solved" is not a very good text, I find. Maybe you can say the reason to guide the user. Maybe something like "due to too many possible solutions" or whatever...

I have  have no information about this. I can stop with timout or loop counter (act state). I have no soltions in this situation.

Mike-3:
> 

grafik
Solution for smal boxes

PR #1271, opened 2023-02-12
Update double_type_utils.dart

t-m-z:
add function

bool isDouble(String text){
  return double.tryParse(text) != null;
}


t-m-z:
will add this as a part of the null-safety-refactoring

PR #1339, opened 2023-04-25
Formula list widget

Mike-3:
widget for formula solver and variable coordinates formulas
add base data type
variable coordinate export / import


S-Man42:
conflicts...

S-Man42:
still conflicts...

lib/common_widgets/gcw_formula_list.dart
```diff
@@ -0,0 +1,283 @@
...
+import 'package:gc_wizard/utils/string_utils.dart';
+
+
+import 'dialogs/gcw_delete_alertdialog.dart';
+import 'dialogs/gcw_dialog.dart';
+import 'gcw_popup_menu.dart';
+import 'gcw_text_export.dart';
+
+
+class GCWFormulaListEditor extends StatefulWidget {
```
S-Man42, line 23: Call the file "_editor" as well.
I would place the widget to Formulas. VarCoords itself uses Formula widgets nonetheless, but no other, so it is ok, to handle that as formula widget... (unless, see further comment on the comment section of the PR)

S-Man42:
VarCoords: There's an "Export" section now. But not yet translated.

S-Man42:
I had a strange behaviour when tested first time: The setState() was not triggered after Adding a new Formula... However, I was not able to reproduce it... Maybe you could have a look, if something's wrong there nonetheless.

S-Man42:
General thought: For me, it is well done and good work at all. But I am thinking, that this kind of widget could be much more general. Because I believe Formula and Varcoords could not be the only tools using it. I believe it could be used for any other as well. Think about complicated inputs of Enigma or whatever. Maybe you want to store different ones. Then you'll need such "Entry" widget. Maybe it could be possible to generalize it a bit (Option for naming the titles, etc.). If we'd ever decide to make a widget storeable, you would need it...

Mike-3:
> General thought: For me, it is well done and good work at all. But I am thinking, that this kind of widget could be much more general. Because I believe Formula and Varcoords could not be the only tools using it. I believe it could be used for any other as well. Think about complicated inputs of Enigma or whatever. Maybe you want to store different ones. Then you'll need such "Entry" widget. Maybe it could be possible to generalize it a bit (Option for naming the titles, etc.). If we'd ever decide to make a widget storeable, you would need it... But in that case, you can't call everthing "formula". Maybe "group" as default. And then you can put in "group_name_singular" and "group_name_plural", so maybe you could write
> 
> i18n('delete') + i18n(widget.group_name_singular)
> 
> -->
> 
> "Delete" + "Formula" or "Delete" + "Group"
> 
> I will merge this as it is, but maybe we should think about making it more general.

I don't think that's a good thing. because in other languages that might not fit the order.

S-Man42:
Well, that was only a quick sketch. Of course, much better is to work with wildcards, of course: 

en
delete_formula = "Delete %s"
delete_group = "Group"

de
delete_formula = "%s wird gelöscht" // jaa, ich weiß, es geht ums Prinzip
delete_group = "Gruppe"
--> 
group = i18n('delete_group')
text = i18n('delete_formula', [group]);
-->
en: Delete Group
de: Gruppe wird gelöscht

PR #1381, opened 2023-05-23
GCWizard Deep links

Mike-3:
first version to discuss und tests
open the feature and configure the widget with parameters
with alphabet values, morse, rotate and multi_decoder


lib/tools/crypto_and_encodings/morse/widget/morse.dart
```diff
@@ -5,17 +5,33 @@ import 'package:gc_wizard/application/theme/theme_colors.dart';
...
 import 'package:gc_wizard/common_widgets/outputs/gcw_output_text.dart';
 import 'package:gc_wizard/common_widgets/switches/gcw_twooptions_switch.dart';
 import 'package:gc_wizard/common_widgets/textfields/gcw_textfield.dart';
 import 'package:gc_wizard/tools/crypto_and_encodings/morse/logic/morse.dart';
 import 'package:gc_wizard/utils/ui_dependent_utils/text_widget_utils.dart';
 
-class Morse extends StatefulWidget {
-  const Morse({Key? key}) : super(key: key);
+const String info ='''
+{
```
S-Man42, line 16: I would like to orientate the definition exactly at the OpenAPI 3 specification:

```
{
	"/morse" : {
		"get": {
			"summary": "Morse Tool",
			"responses": {
				"204": {
					"description": "Tool loaded. No response data."
				}
			}
		},
		"parameters" : [
			{
				"in": "query",
				"name": "input",
				"required": true,
				"description": "Input data for encoding or decoding Morse",
				"schema": {
					"type": "string"
				}
			},
			{
				"in": "query",
				"name": "mode",
				"description": "Defines encoding or decoding mode",
				"schema": {
					"type": "string",
					"enum": [
						"encode",
						"decode"
					],
					"default": "encode"
				}
			}
		]
	}
}
```

This makes it easier for client to identify and parse the fields without having to learn new syntaxes. And this avoids that every tool api designer creates its own api definition.

lib/tools/crypto_and_encodings/morse/widget/morse.dart
```diff
@@ -5,17 +5,33 @@ import 'package:gc_wizard/application/theme/theme_colors.dart';
...
+  "parameters": [
+    "inputParam": {
+      "name": "input",
+      "description": "input value",
+    },
+    "modeencodeParam": {
```
S-Man42, line 22: I'd simply call it "mode"

lib/application/main_menu/deep_link.dart
```diff
@@ -0,0 +1,215 @@
...
+        return GCWOutputText(
+            text: snapshot.data ?? '',
+            copyText: 'https://test.gcwizard.net/#/' + tool.id
+        );
+      }
+  );
+}
+
+Future<String> _toolInfoText(GCWTool tool) async {
+  var info = tool.id;
```
S-Man42, line 179: simply add the OpenAPI Spec in a GCWCodeView dingens

lib/tools/crypto_and_encodings/alphabet_values/widget/alphabet_values.dart
```diff
@@ -22,16 +23,32 @@ import 'package:gc_wizard/tools/formula_solver/persistence/model.dart';
...
 import 'package:gc_wizard/utils/collection_utils.dart';
+import 'package:gc_wizard/utils/complex_return_types.dart';
 import 'package:gc_wizard/utils/constants.dart';
 import 'package:gc_wizard/utils/data_type_utils/object_type_utils.dart';
 import 'package:gc_wizard/utils/json_utils.dart';
 import 'package:prefs/prefs.dart';
 
-class AlphabetValues extends StatefulWidget {
-  const AlphabetValues({Key? key}) : super(key: key);
+const String info ='''
```
S-Man42, line 32: Call it "apiSpecification"

S-Man42: Is there a way to force this text? I mean, maybe the GCWWebStatefulWidget could be abstract. And it could have a

```
@abstract
String apiSpecification() {}
```

So, that implementing classes are forced to fill out this field.

lib/tools/crypto_and_encodings/alphabet_values/widget/alphabet_values.dart
```diff
@@ -61,6 +78,21 @@ class _AlphabetValuesState extends State<AlphabetValues> {
   void initState() {
     super.initState();
 
+    if (widget.hasWebParameter()) {
```
S-Man42, line 81: somehow I don't like that this kind of code block must be added to all the WebTools from now on... Maybe a better, more generic way? (I have no idea by myself in the moment, but one should think about....)

lib/tools/crypto_and_encodings/alphabet_values/widget/alphabet_values.dart
Mike-3: is a parameter
```
  final String? apiSpecification;

  GCWWebStatefulWidget({Key? key, this.webParameter, required this.apiSpecification}): super(key: key);
```

PR #1397, opened 2023-06-02
Tmz 3.0.0 enhance base64

t-m-z:
add feature to differ between text, image and sound base64 encoded data and show it


S-Man42:
Could you please add a test?

t-m-z:
added

S-Man42:
We should definitely add a Waiting Spinner when input gets too long. My emulator froze at the first example I tried... a 316kb JPG, 1440x960px

t-m-z:
AsyncExecuter added - but this will not solve the problm with the clipboard

PR #1398, opened 2023-06-02
Tmz 3.0.0 enhance keyboard - closes #1367

### LINKED ISSUE #1367: Feature request: keyboard characters to numbers
Hi,

Is it possible to have a keyboard to numbers converter (QWERTY keyboard characters/number keys) as well as international keyboard differences to numbers/letters)

Geocache example: GC60XQJ

Thanks

James A

t-m-z:
add two maps: iPhone and Android
Geocache example: GC60XQJ


S-Man42:
Android is extremely different concerning its keyboards. For example my Swiftkey has no special chars at the numbers, they all lie above the letters.

I am quite sure, that Motorola differs from Samsung from Xiaomii from...

So, it would be necessary to write which model you used for this. Maybe iPhone versions differ as wekk?

t-m-z:
The challenge are the company specific changes to the pure google android.
While many companies stick to google like Motorola or OnePlus others like samsung are extremely fancy.
I will check and find a way

Sent from MailDroid

-----Original Message-----

PR #1400, opened 2023-06-02
Tmz 3.0.0 gcd lcm - closes #1203

### LINKED ISSUE #1203: feature: GGT, KGV

t-m-z:
greatest common divisor - ggT
least commen multiplier - kgV


lib/tools/science_and_technology/gcd_lcm/logic/gcd_lcm.dart
```diff
@@ -0,0 +1,20 @@
+import 'dart:core';
+
+int gcd(int a, int b) {
```
S-Man42, line 3: For me functions for the common math lib.

test/tools/science_and_technology/gcd_lcm/logic/gcd_lcm_test.dart
```diff
@@ -0,0 +1,39 @@
+import "package:flutter_test/flutter_test.dart";
+import 'package:gc_wizard/tools/science_and_technology/gcd_lcm/logic/gcd_lcm.dart';
+
+void main() {
+  group("gcd", () {
+    List<Map<String, Object?>> _inputsToExpected = [
+      {'input_a' : 0, 'input_b' : 0, 'expectedOutput' : 0},
```
S-Man42, line 7: What about negative values?

lib/tools/science_and_technology/gcd_lcm/widget/gcd_lcm.dart
```diff
@@ -0,0 +1,66 @@
+import 'package:flutter/material.dart';
+import 'package:gc_wizard/application/i18n/app_localizations.dart';
+import 'package:gc_wizard/common_widgets/outputs/gcw_columned_multiline_output.dart';
+import 'package:gc_wizard/common_widgets/outputs/gcw_default_output.dart';
+import 'package:gc_wizard/common_widgets/spinners/gcw_integer_spinner.dart';
+import 'package:gc_wizard/tools/science_and_technology/gcd_lcm/logic/GCD_LCM.dart';
+
+
+class GCDLCM extends StatefulWidget {
```
S-Man42, line 9: I'd prefer two different tools for two different functions

t-m-z:
reafctor done

assets/searchstrings/common.json
```diff
@@ -175,6 +175,7 @@
   "games_scrabble": "scrabble",
   "games_sudokusolver": "sudoku",
   "gccode": "geocachingcode geocachecode gccode base31",
+  "gcd_lcm": "greatestcommondivision gcd leastvommonmultiple lcm",
```
S-Man42, line 178: This is not for common, this has to be translated

S-Man42:
grafik
Why are you using a list output widget, not a normal one? You don't have a list, don't you?

S-Man42:
grafik
For some reason, not clear to me at all, i can't find anything when searching for lcm

S-Man42:
Currently you are returning 0 when an input number is 0. But in my opinion, thats wrong, isn't it. 0 can't be a divider any time because that would give a division by zero. So in my opinion, to be correct. That's true for multiple. 0 can't be a multiple of anything. Because that would mean, this number could be divided by 0 as well. So 0 is no valid input for LCM or GCD. Maybe here a message would be more correct, what do you think?

assets/searchstrings/en.json
```diff
@@ -240,6 +241,7 @@
   "languagegames_piglatin": "piglatin",
   "languagegames_robberlanguage": "robberlanguage",
   "languagegames_spoonlanguage": "spoonlanguage spoonspeak",
+  "lcd": "least common multiplier",
```
S-Man42, line 244: lcd? isn't it lcm? and here the searchstring lcm is missing

t-m-z:
> Currently you are returning 0 when an input number is 0. But in my opinion, thats wrong, isn't it. 0 can't be a divider any time because that would give a division by zero. So in my opinion, to be correct. That's true for multiple. 0 can't be a multiple of anything. Because that would mean, this number could be divided by 0 as well. So 0 is no valid input for LCM or GCD. Maybe here a message would be more correct, what do you think?

according to Wikipedia: 
for the case m=0 or n=0 is the lcm definded as lcm(m; n) = 0
Harald Scheid: Einführung in die Zahlentheorie. Klett Verlag, Stuttgart, 1972, [ISBN 3-12-983240-8](https://de.wikipedia.org/wiki/Spezial:ISBN-Suche/3129832408), S. 79.

t-m-z:
> Currently you are returning 0 when an input number is 0. But in my opinion, thats wrong, isn't it. 0 can't be a divider any time because that would give a division by zero. So in my opinion, to be correct. That's true for multiple. 0 can't be a multiple of anything. Because that would mean, this number could be divided by 0 as well. So 0 is no valid input for LCM or GCD. Maybe here a message would be more correct, what do you think?

according to  https://www.matheboard.de/archive/592607/thread.html
ggt(0,0) is defined as 0

t-m-z:
searchstrings corrected,
widget refactored
regarding 0 - these are defined by number theory

PR #1402, opened 2023-06-02
Tmz 3.0.0 spelling alphabets - closes #1227, 1359

### LINKED ISSUE #1227: List of Spelling Alphabets
https://en.wikipedia.org/wiki/Spelling_alphabet

t-m-z:


S-Man42:
1. I'd like the take the "Germany" instead of "Germany 2022" as default. It is much more common.
2. Here we have the same thing as already discussed: The spaces for word separation. I see your point, but in my opinion we should avoid this (And why are there 3 spaces?): If the code does not specify an encoding for any character, we should ignore it. As in any other encryption.
3. Some sort orders are quite strange. For example Polish.
4. Either put the overview page into the search engine or the subtools, but not both. We decided to don't do that to avoid too much and too confusing search results. Which of both variants differ and depend from case to case. Here I guess we should take the overview, but I am not quite sure about it.

t-m-z:
Valid points.
Regarding the default I would prefer NATO ;-)
You are right that Germany is much more common but its officially deprecated.

To avoid the whitespace discussion the substitution tool could be the answer. I will check this.

I will also think about the overview thing and compare it to the other tools again.

Sent from MailDroid

-----Original Message-----

S-Man42:
Concerning the spaces:
No, I dont believe that the subst tool is the answer. Its a question of consistency from my point of view. And the GCW works as this. It feels extremely... willkürlich... adding 3 (!) spaces between words. In most other tools unknown characters are not unconsidered. Either we decide to add word delimiters to all tools (i'd not like this because if not defined, ignore them imho) or throw them away.

I talked to @Mike-3 on this, as I underatood him, he's  the same opinion...

I see your point, I understand why you do this. But really, I strongly recommend to discuss this point globally. It must be something really consistent. Otherwise every tool behaves different and thats very confusing.

PR #1415, opened 2023-06-11
 enhance wasd to draw diagonal

t-m-z:


lib/tools/crypto_and_encodings/wasd/logic/wasd.dart
```diff
@@ -181,14 +193,18 @@ String decodeWASD(String input, List<String> controlSet) {
...
-  while (normalized.length < 4) {
+  while (normalized.length < 9) {
     normalized.add(null);
   }
 
   if (normalized[0] == null || normalized[0]!.isEmpty) normalized[0] = '↑';
   if (normalized[1] == null || normalized[1]!.isEmpty) normalized[1] = '←';
   if (normalized[2] == null || normalized[2]!.isEmpty) normalized[2] = '↓';
   if (normalized[3] == null || normalized[3]!.isEmpty) normalized[3] = '→';
+  if (normalized[4] == null || normalized[4]!.isEmpty) normalized[4] = '↖';
```
S-Man42, line 204: It seems strange to me to find the cursors that often in the code. If one needs to change them, he needs to touch several different places in the code. It would be better to make it: 

make KeyBoardControls a Class. So you could access the attributes left, right, topleft, etc.

Then 

Cursors = KeyBoardControls(left: ->, right: <-, ...)

DEFAULT_SET = Cursors, 
access DEFAULT_SET.left whereever you need it.

t-m-z: you are right - however, as I remember, it was your code.
I believe there is no need to change them

As far as I understand the code every digit is encoded with a series of characters which move a virtual pen in a specific direction and draw this digit.
These arrows are the standard set which where substituted with the standard set

S-Man42: I don't know who wrote this, maybe it was me. But it is a code smell. There is a scout rule for software engineers: If you find trash, remove it. Means, if there's something smelly, we should fix it when we come by. This one looks smelly: Using special characters several times in the code as identifiers should be fixed imho

t-m-z:
Hm. Okay. 
Probably it will be a solution to define a const for every direction like CURSOR_UP ...  and replace every special charcter with its const.

Sent from MailDroid

-----Original Message-----

S-Man42:
But then dont forget to use these consts in the WASDTYPE.CURSORS Definition as well :)

S-Man42:
kümmere mich drum.

t-m-z:
Herzlichen Dank. Hilft mir in meiner Sauna auf dem Dachboden 

Sent from MailDroid

-----Original Message-----

PR #1421, opened 2023-06-17
feature unix time. excel time #closes 1420

t-m-z:
feature wish from Nasenbaer69


t-m-z:
yes I thought about it. 

Regarding https://en.wikipedia.org/wiki/Unix_time Unix time by design does not require a specific size for the storage. Hence there will be no problem in theory.
Unfortunately in the real world sometimes 32bit integers are used to store the time. Hence this year 2038 Problem will arise - https://en.wikipedia.org/wiki/Year_2038_problem
Therefore a modulo and recount could be an option.

S-Man42:
I believe, JS could make problems nonetheless. We already had this in the past: While dart does not have any issues with numbers in that size, JS has a limitation (not quite sure which). But we should have an eye on it. I believe, it's such a crazy thought that somebody may build a computer-based mystery where he might use this range border for something. Not in 12 years. So we should think about how to face it.

I will merge it, but maybe you'll get an idea to deal with it.

S-Man42:
Please add tests

t-m-z:
tests added

S-Man42:
grafik

Irgendwo packt der Sekunden-Nachkommastellen aus. Bitte mal checken, wo das herkommt. Fraglich ist überhaupt, ob Nachkommastellen überhaupt Sinn ergeben.

S-Man42:
Encryption: Maybe it could be a good idee to add a Divider with (year - month -day) desciption. US amis may switch day and month or something.

Decryption: Please return the localized date.

S-Man42:
Negative Unix Time in "Calender"? Does this make any sense? If so, we should be consistent and add it to the UNIX Timestamp Tool as well (in that case please add tests just to demonstrate the feature)

grafik

assets/searchstrings/de.json
```diff
@@ -127,7 +127,7 @@
   "dates_day_of_the_year": "tagesnummern tagnummern tagzahlen tageszahlen tagesanzahlen wochentage wochennummern wochenzahlen wochenanzahlen jahrestage tagimjahr tagdesjahres",
   "dates_timecalculator": "uhrzeit uhrzeitrechner stunden minuten sekunden zeitenrechner",
   "dates_weekday": "wochentage",
-  "dates_calendar": "julianisches datum gregorianisches datum kalender",
+  "dates_calendar": "julianisches datum gregorianisches datum kalender zeitstempel maya koptisch potrzebie tzolkin haab yazdegard",
```
S-Man42, line 130: As usual, translations only in Crowdin! Otherwise we might get conflicts.

t-m-z:
> grafik
> 
> Irgendwo packt der Sekunden-Nachkommastellen aus. Bitte mal checken, wo das herkommt. Fraglich ist überhaupt, ob Nachkommastellen überhaupt Sinn ergeben.

will be removed with localized date format

t-m-z:
> Encryption: Maybe it could be a good idee to add a Divider with (year - month -day) desciption. US amis may switch day and month or something.
> 
> Decryption: Please return the localized date.

Encyption: this means to enhance the DateTimePicker-Widget according to locale settings

Encryption: done

S-Man42:
"Encyption: this means to enhance the DateTimePicker-Widget according to locale settings"

Dont understand. Whats the problem by simply add a divider widget before the datepicker. There's no reason to touch the date picker...

t-m-z:
> "Encyption: this means to enhance the DateTimePicker-Widget according to locale settings"
> 
> Dont understand. Whats the problem by simply add a divider widget before the datepicker. There's no reason to touch the date picker...

Your point ouches every usage of the DateTimePicker - should I really add a Divider before every usage? W used this widget about 14 times

S-Man42:
hmm... ok we shoukd think about more generel

t-m-z:
done - however I see a strange behaviour with eth Excel-Timestamp. There is a kind of overflow Error I can not really identify.

S-Man42:
> There is a kind of overflow Error I can not really identify.

Where is it? What needs to be done? Maybe leave Excel out for the moment and only insert Unix?

S-Man42:
Please make "Decrypt" mode the initial one.

PR #1424, opened 2023-06-23
use screen height from theme

Mike-3:
sudoku
grid
game of life


S-Man42:
I believe the *0.8 was quite relevant for layout.

PR #1434, opened 2023-07-03
bugfix formula editor

t-m-z:


t-m-z:
formular editor - exception when accessing a null value
hexviewer - missing widgets according to 2.3.1

lib/tools/formula_solver/widget/formula_solver_values.dart
```diff
@@ -37,7 +37,7 @@ class _FormulaSolverFormulaValuesState extends State<_FormulaSolverFormulaValues
       maxLetterIndex = max(maxLetterIndex, alphabetIndex);
     }
 
-    if (alphabet_AZIndexes.keys.contains(maxLetterIndex)) {
+    if (alphabet_AZIndexes.keys.contains(maxLetterIndex + 1)) {
```
S-Man42, line 40: I am not quite sure if this fixes the error. Did you check with a 2.x code version? I believe there's a more structural issue which effects the second bug with the non-visible A in the values. I woulf like to check it a more deeper to ensure not just doing something on symptomes.

Primary check would be: Is there a +1 in the old code?

t-m-z: the code is rewritten. The old code says
```
if (maxLetterIndex < 26) {
      return alphabet_AZIndexes[maxLetterIndex + 1];
    }
```

I assume this similar to the +1

Issue #1253, opened 2023-01-31
Refactoring: NavigationService vs. NoAnimationMaterialPageRoute


S-Man42:
Currently it is not quite clear to me whether both classes do the same or not.

Usually we are using second to open a tool. In just 2 cases we use the NavigationService (Main and GCWPasteButton).

Is it possible to get rid of one of both?

PR #1437, opened 2023-07-06
MaterialPageRoute -> NoAnimationMaterialPageRoute

Mike-3:
null problem with clipboard editor persists


S-Man42:
> null problem with clipboard editor persists

What does this mean. Does this break anything?

S-Man42:
This does not affect the original ticket #1253 which is not about "MaterialPageRoute" but "NavigationService"

Mike-3:
> What does this mean. Does this break anything?

No, everything is just the same as before.

Mike-3:
> This does not affect the original ticket #1253 which is not about "MaterialPageRoute" but "NavigationService"

I thought that meant both

S-Man42:
It's ok what you did, indeed. However, the more important one is the NavigationService which is used at one part instead of the RouteThing

PR #1446, opened 2023-07-10
Consolidate Load and Save Functions 

Mike-3:
convert string to bytes
only 1 export file routine

### DIFF
--- a/lib/utils/ui_dependent_utils/file_widget_utils.dart
+++ b/lib/utils/ui_dependent_utils/file_widget_utils.dart
@@ -11,41 +11,16 @@ import 'package:gc_wizard/utils/file_utils/file_utils.dart';
 import 'package:universal_html/html.dart' as html;
 
 Future<bool> saveByteDataToFile(BuildContext context, Uint8List data, String fileName) async {
-  if (kIsWeb) {
-    var blob = html.Blob([data], 'image/png');
-    html.AnchorElement(
-      href: html.Url.createObjectUrl(blob),
-    )
-      ..setAttribute("download", fileName)
-      ..click();
-
-    return Future.value(true);
-  } else {
-    var storagePermission = await checkStoragePermission();
-    if (!storagePermission) {
-      showToast(i18n(context, 'common_exportfile_nowritepermission'));
-      return false;
-    }
-
-    fileName = _limitFileNameLength(fileName);
-    final fileInfo = await FilePickerWritable().openFileForCreate(
-      fileName: fileName,
-      writer: (file) async {
-        await file.writeAsBytes(data);
-      },
-    );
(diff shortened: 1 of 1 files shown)


S-Man42:
Interesting is still that the load routine is still untouched. So, the back-operation still does not reconvert the loaded bytes correctly into a String. convertBytesToString is never called, isnt it?

Mike-3:
We create only GCW files with byte data. So it has to be called individually every time it is converted to a string.

S-Man42:
this makes sense

lib/utils/ui_dependent_utils/file_widget_utils.dart
```diff
@@ -11,15 +11,23 @@ import 'package:gc_wizard/utils/file_utils/file_utils.dart';
...
+}
+
+Future<bool> saveStringToFile(BuildContext context, String data, String fileName) async {
+  return _saveDataToFile(context, convertStringToBytes(data), fileName, true);
+}
+
+Future<bool> _saveDataToFile(BuildContext context, Uint8List data, String fileName, bool byteData) async {
   if (kIsWeb) {
-    var blob = html.Blob([data], 'image/png');
+    var blob = html.Blob([data], byteData ? 'image/png' : 'text/plain', byteData ? null : 'native');
```
S-Man42, line 23: Are you quite sure that this really works? So you say: byteData IS ALWAYS an PNG image. But we do save any other file type as well (hidden data, wig analyzer...). Do I miss something?

Mike-3:
I have seen too.
That's how it was before and we never had any problems with it, right?
I also don't know what should change when saving files if we specify other types there.

S-Man42:
The the MIME Type has to be taken from the GCWFile File Type.

lib/utils/ui_dependent_utils/file_widget_utils.dart
```diff
@@ -20,7 +20,11 @@ Future<bool> saveStringToFile(BuildContext context, String data, String fileName
 
 Future<bool> _saveDataToFile(BuildContext context, Uint8List data, String fileName, bool byteData) async {
   if (kIsWeb) {
-    var blob = html.Blob([data], byteData ? 'image/png' : 'text/plain', byteData ? null : 'native');
+    var fileType = fileTypeByFilename(fileName);
+    String? mimeType;
```
S-Man42, line 24: What happens on fileType == null?

Mike-3:
No problem. The parameter is optional.

S-Man42:
then saveStringToFile should add text/plain?

Mike-3:
Si

S-Man42:
Ok :D Do you add this? :) :)

PR #1451, opened 2023-07-14
grid scale

Mike-3:


S-Man42:
As I see it, you added the same code to all four tools. This seems to be a new GCW Widget, don't you think? A GCWPainterContainer or something...

Mike-3:
ok but please merge it then it's easier for me to change in all. Also hits the pyramid.

PR #1464, opened 2023-07-16
symbol table blue monday code - closes #1238 #1458 #1461

### LINKED ISSUE #1238: Blue Monday Code
https://kryptografie.de/kryptografie/chiffre/blue-monday.htm

t-m-z:


pubspec.yaml
```diff
@@ -85,6 +85,7 @@ flutter:
   assets:
     - assets/logo/
     - assets/tokens/
+    - assets/searchstrings/
```
S-Man42, line 88: Great, you already worked with the new asset structure! However, one little thing remained: Please remove this entry!

lib/application/searchstrings/assets/common.json
```diff
@@ -379,6 +379,7 @@
   "symbol_baudot": "Jeanmauriceemilebaudot",
   "symbol_birds_on_a_wire": "",
   "symbol_blox": "blox",
+  "symbol_blue_monday": "bluemonday neworder petersavilles",
```
S-Man42, line 382: maybe you find some more descriptive string? squares, rectangles, boxes, ...(isnt there already a search string only for squares and rectangles?)

pubspec.yaml
t-m-z: done

lib/application/searchstrings/assets/common.json
t-m-z: yes - but this belong to en-searchstrings - not common   ;-)

t-m-z:
One problem: I did not find a suitable unicode for greek small letter epsilon and gamma

lib/application/searchstrings/assets/common.json
S-Man42: correct

lib/application/main_menu/licenses.dart
```diff
@@ -107,6 +107,8 @@ class _LicensesState extends State<Licenses> {
           const ['several', 'myGeoTools'],
           const ['several', 'Wikipedia'],
           [i18n(context, 'symboltables_alien_mushrooms_title'), '(Personal Use)'],
+          [i18n(context, 'symboltables_berber_title'), 'https://en.wikipedia.org/wiki/Tifinagh (Wiki Commons)'],
```
S-Man42, line 110: Please always add different licenses also to the license file... maybe we need a better way for this to avoid redundant information points, but currently it is like that...

t-m-z: Sure - but where do I find this licence file?

lib/application/searchstrings/assets/common.json
```diff
@@ -568,7 +570,7 @@
   "symbol_terzi": "francesco lana de terzi",
   "symbol_theban": "onorius wikka wicca wicka",
   "symbol_three_squares": "3",
-  "symbol_tifinagh": "tifinagh",
+  "symbol_tifinagh": "tifinagh berber",
```
S-Man42, line 573: i believe i already added berber to en searchstrings. i believe there are lamguages, where "berber" is not "berber" but... "berbres" or somethung. thats why i think, it's not common... it is only the same in en and de by chance...

lib/application/main_menu/licenses.dart
S-Man42: search for LICENSE ;)

S-Man42: As in every open sourcr project it's in thr main root directory ;) https://github.com/S-Man42/GCWizard/blob/3.0.0/LICENSE

lib/application/searchstrings/assets/common.json
t-m-z: Ups. Yes . But Berber is used in ENG and DEU - it should then be added to common

S-Man42: NO!!,

S-Man42: You ALREADY added into common at the moment. But that's my point: Please remove it from there! Only because it is the same in DE and EN, it does not mean that it is valid for all languages. I believe there are languages, where berber is not the correct term. so it is language specific.

t-m-z: Hm. Still not convinced. But okay

PR #1466, opened 2023-07-18
symbol table interlac

Mike-3:


S-Man42:
Please think about searchstrings: It is even no string vor "interlac", if I see correctly. Furthermore: DC detectivecomics legionofsuper-heroes (notice the hyphen!) aliens

special adding: sheldoncooper thebigbangtheory greenlantern (because he wears a shirt with these signs extremely often printed on a green lantern logo)

S-Man42:
conflicts

lib/application/searchstrings/assets/en.json
```diff
@@ -474,7 +474,7 @@
   "symbol_illuminati_v1": "malesign",
   "symbol_illuminati_v2": "femalesign asterisk triangles rectangles hairpins bobbypins circles hourglassshape",
   "symbol_intergalactic": "intergalactical galaxy futuristic alien dots stripes thicklines angularsigns",
-  "symbol_interlac": "",
+  "symbol_interlac": "dc detectivecomics legionofsuper-heroes aliens",
```
S-Man42, line 477: i'd add both versions: legionofsuper-heroes and legionofsuperheroes. It's because of both versions are possible searchstrings.... maybe we should think about removing hyphens before search.... don't you think?

PR #1479, opened 2023-08-04
Update about.dart

t-m-z:
hide twitter


S-Man42:
Good idea! Remove it completely!

Also from en.json (other languages will be eliminated autmatically by Crowdin afaik)

PR #1482, opened 2023-08-05
tmz 3.0.0 catch microsoft excel bug and enhance excel timestamp

t-m-z:
https://learn.microsoft.com/en-us/office/troubleshoot/excel/wrongly-assumes-1900-is-leap-year


lib/application/i18n/assets/en.json
```diff
@@ -19,6 +19,9 @@
   "about_manual": "Online manual",
   "about_manual_url": "https://blog.gcwizard.net/manual/en/",
   "about_manual_url_text": "blog.gcwizard.net/manual/en/",
+  "about_twitter": "Twitter",
```
S-Man42, line 22: again....

lib/tools/science_and_technology/date_and_time/calendar/widget/calendar.dart
```diff
@@ -71,7 +72,8 @@ class _CalendarState extends State<Calendar> {
           GCWIntegerSpinner(
               value: _currentTimeStamp,
               min: 0,
-              max: 8640000000000, //max days in seconds according to DateTime https://stackoverflow.com/questions/67144785/flutter-dart-datetime-max-min-value
+              max:
+                  8640000000000, //max days in seconds according to DateTime https://stackoverflow.com/questions/67144785/flutter-dart-datetime-max-min-value
```
S-Man42, line 76: phew... i guess this could be a problem at least in the online version, because JavaScript has Problems with big numbers...

lib/tools/science_and_technology/date_and_time/calendar/logic/calendar.dart
```diff
@@ -43,11 +43,13 @@ String JulianDateToUnixTimestamp(double jd) {
 }
 
 double ExcelTimestampToJulianDate(int timestamp) {
+  Duration days = Duration(days: timestamp);
+  if (timestamp > 60) days = days - const Duration(days: 1); // correct Excel Bug - 1900 is a leap year
```
S-Man42, line 47: I dont get this: You always subtract one day if it is more than 60 days???

S-Man42: What has this to do with year 1900?

lib/tools/science_and_technology/date_and_time/calendar/widget/calendar.dart
S-Man42: isnt that even bigger than our MAX_INT variable?

t-m-z:
Microsoft assumes that 1900 is a leap year because of compatibility reasons to Lotus 124. It is wiered but known for 35 years.
Hence according to Microsoft the 29.02.1900 exists and is counted.
So there is a difference of one day after the 1st of March.

Sent from MailDroid

-----Original Message-----

lib/application/i18n/assets/en.json
t-m-z: is done with a new pull from 3.0.0

lib/tools/science_and_technology/date_and_time/calendar/widget/calendar.dart
t-m-z: There should be no problem

https://stackoverflow.com/questions/31687376/what-is-the-max-value-of-integer-in-dart

That depends if you are running in the Dart VM or compiling to JavaScript. 
On the Dart VM an int is arbitrary precision and has no limit.
When compiling to JavaScript you are actually using floating point numbers, since that is all that JavaScript supports. This means you are restricted to representing integers within the range -2^53 to 2^53

2^53 = 9,007,199,254,740,992 > 8,640,000,000,000

lib/tools/science_and_technology/date_and_time/calendar/logic/calendar.dart
t-m-z: Microsoft assumes that 1900 is a leap year because of compatibility reasons to Lotus 124. It is wiered but known for 35 years.
Hence according to Microsoft the 29.02.1900 exists and is counted.
So there is a difference of one day after the 1st of March.

https://learn.microsoft.com/en-us/office/troubleshoot/excel/wrongly-assumes-1900-is-leap-year

lib/tools/science_and_technology/date_and_time/calendar/widget/calendar.dart
S-Man42: No, in pure Dart it is no problem, but we already had massive problems with the JS web version.

Please compile for web and check

lib/tools/science_and_technology/date_and_time/calendar/logic/calendar.dart
S-Man42: Yes, I understand the line after. But this special one with the 60 days is not clear to me.

t-m-z:
Its because of the way Excel calculates. The Timestamp 60 points to 29.02.1900. Hence from that day forward the excel calcutes one day plus.
Excel counts days. Hour, min an sec are stored in the fractional part of the Timestamp 

Sent from MailDroid

-----Original Message-----

PR #1484, opened 2023-08-08
Geohashing 

Mike-3:


lib/application/searchstrings/assets/common.json
```diff
@@ -177,6 +177,7 @@
   "games_scrabble": "scrabble",
   "games_sudokusolver": "sudoku",
   "gccode": "geocachingcode geocachecode gccode base31",
+  "geohashing": "geohashing dowjonesindex",
```
S-Man42, line 180: xkcd boersenkurs (no idea what it means in english) money dates days locations....

Please invest much more time into search strings

lib/common_widgets/coordinates/gcw_coords/coord_format_inputs/degrees_latlon/degrees_lat_textinputformatter.dart
```diff
@@ -1,9 +1,9 @@
-part of 'package:gc_wizard/common_widgets/coordinates/gcw_coords/gcw_coords.dart';
+import 'package:flutter/services.dart';
```
S-Man42, line 1: is this necessary?

S-Man42:
I am not able to read everythin atm. Only a thought: I guess you embedded an API url somewhere. Are you quite sure that this one will be available all the time? I thought maybe a good idea to add two or three different (if available, I really dont know). So, the code automatically switches if the first may be broken and we have time to react.

And we should implement something, maybe on Thomas' server where we get an alarm if onr URL may be offline so that we can react.

lib/common_widgets/coordinates/gcw_coords/coord_format_inputs/degrees_latlon/degrees_lat_textinputformatter.dart
Mike-3: Since I also wanted to use the InputFormater for geohashing, I had to adapt it.

Mike-3:
I can look for alternate site with rest api for DOW. But actually the main function should be offline.

S-Man42:
> I can look for alternate site with rest api for DOW. But actually the main function should be offline.

Of course. But it should be resistant to external changes.

lib/common_widgets/coordinates/gcw_coords/coord_format_inputs/degrees_latlon/degrees_lat_textinputformatter.dart
S-Man42: hmm... ok...

S-Man42:
yes, not for this small tool. Ok, but let us think about an alarm system in case of website down.

t-m-z:
If it is a user specific token it might be a solution to provide preference key where the user could store his token.
If there is none gehashing will not be used

Sent from MailDroid

-----Original Message-----

Mike-3:
Actually, I have integrated an API where no token is required. I think that's enough for now.

lib/tools/coords/geohashing/widget/geohashing.dart
```diff
@@ -0,0 +1,302 @@
...
+    _monthController.dispose();
+    _dayController.dispose();
+
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Column(children: <Widget>[
+      GCWTextDivider(text: '', trailing: _buildTrailingButtons(IconButtonSize.SMALL)),
```
S-Man42, line 89: Please add description: "Year, Month, Day"

lib/application/i18n/assets/en.json
```diff
@@ -2186,6 +2186,11 @@
   "gcd_gcd": "Greatest Common Divisor",
   "generalcodebreakers_selection_title": "General Code Solvers",
   "generalcodebreakers_selection_description": "Tools, which try to identify or break codes without any further information (Multi Decoder, Hash Search, ...)",
+  "geohashing_title": "Geohashing",
+  "geohashing_description": "Coordinate format from xkcd-Comic. Format: 2023-09-10 34 -111",
```
S-Man42, line 2190: Remove Format hint here, it's better directly at the widget. 

Furthermore: "xkcd webcomic" instead of "xkcd-comic".

linux/flutter/ephemeral/.plugin_symlinks/audioplayers_linux
```diff
@@ -0,0 +1 @@
+C:/Users/MikeB/AppData/Local/Pub/Cache/hosted/pub.dev/audioplayers_linux-3.0.0/
```
S-Man42, line 1: do not add the linux crap here!

lib/tools/coords/geohashing/widget/geohashing.dart
```diff
@@ -0,0 +1,302 @@
...
+      outputs: _currentOutput,
+      points: _currentMapPoints,
+    );
+    Widget? extendedOutput;
+
+    if (_geohashing != null) {
+      var rows = [
+        [i18n(context, 'geohashing_dow_jones_index'),
+          (_geohashing!.dowJonesIndex > 0) ? _geohashing!.dowJonesIndex.toString() : ''],
+        [i18n(context, 'geohashing_title'), _geohashing.toString()],
```
S-Man42, line 185: what is the geohashing string telling the user? I just see the unformatted date followed by two 0 characters

Mike-3: This is the format from geohashing. You didn't specify lat lon, hence the 0.

lib/application/i18n/assets/en.json
Mike-3: The thought was that otherwise hardly anyone would recognize that it was the format they were looking for. It is not taken into account with the normal coordinate formats.

S-Man42: Tbh, I dont understand this, but if you mean, it makes sense... But then this would be better as example, not description

linux/flutter/ephemeral/.plugin_symlinks/audioplayers_linux
Mike-3: Ok, aber wir sollten die dann mal in den 3.0.0 Branch übernehmen oder das git ignore anpassen. Es nervt einfach, die Dateien immer wieder beim einchecken abzuwählen.

lib/tools/coords/geohashing/widget/geohashing.dart
```diff
@@ -138,7 +139,7 @@ class _GeohashingState extends State<Geohashing> {
           setState(() {
             _currentOnline = value;
             if (_currentOnline == GCWSwitchPosition.right) {
-              _showOnlineToast();
+              _showOnlineDialog();
```
S-Man42, line 142: Here is a problem. If the user presses no or cancel, the original value needs to be set. So more correct would be to include the entire setState(_currentOnline = true;) code into the onOkPressed hook of the dialog.

As I see it, at the moment the user can click no, the switch is toggled anyway.

PR #1486, opened 2023-08-14
tmz 3.0.0 featurewish gade to formulaeditor

t-m-z:
add feature wish from Magnus, E-Mail dated from Mo., 14. Aug. 2023, 00:00

Hallo,

Ich habe eben das Feature Gade (Ziffern sortiert, ergänzen, und dann den Buchstaben zuweisen) gefunden.
Danke! Leider habe ich erst entdeckt als ich den Cache GCABXMG schon manuell rausgefummelt hatte.

Ich habe einen Ergänzungsvorschlag:
Unten an der Tabelle sollte auch zwei Kopier- buttons sein.
Als Beispiel
GZWizard 7 3 23 9 26 1 18 4
Gade:  11223346789

Einer für den GC wizard Formelrechner:
Wenn man die Tabelle kopiert sollte also in der Zwischenablage stehen:

["{\"key\":\"A\",\"value\":\"1\"}","{\"key\":\"B\",\"value\":\"1\"}","{\"key\":\"C\",\"value\":\"2\"}","{\"key\":\"D\",\"value\":\"2\"}","{\"key\":\"E\",\"value\":\"3\"}","{\"key\":\"F\",\"value\":\"3\"}","{\"key\":\"G\",\"value\":\"4\"}","{\"key\":\"H\",\"value\":\"6\"}","{\"key\":\"I\",\"value\":\"7\"}","{\"key\":\"J\",\"value\":\"8\"}","{\"key\":\"K\",\"value\":\"9\"}","{\"key\":\"L\",\"value\":\"0\"}","{\"key\":\"M\",\"value\":\"5\"}"]

Und ich fände es toll wenn es noch einen gäbe der das im cgeo Format kopiert:
Wenn man die Tabelle kopiert sollte in der Zwischenablage stehen:
$A=1 | $B=1 | $C=2 | $D=2 | $E=3 | $F=3 | $G=4 | $H=6 | $I=7 | $J=8 | $K=9 | $L=0 | $M=5
Das könnte man dann direkt in die Final-Formel von cgeo kopieren.

Wenn das realisiert wird lege ich einen Cache dazu und verweise auf Euer tolles Tool!

Gruß Magnus


S-Man42:
44 files changes for that Feature? is that correct?

S-Man42:
Hm.... I am not quite lucky with the result...
1.) The c:geo thing... Well, we should add a bit more evaluation on this. I like the idea in general, but not in that way. We should add some support to cgeo, yes, but we should add this functionality in a more central way. For instance, we could add a menu point the formula exporter to export cgeo valid formulas. In that manner we would be able to export any formula we created, not that from the Gade function.
2.) The export to our own system using the clipboard is not a lucky way, imho, too. We should be able to create a new formula group directly and open the tool as following step. Same as we can do from the Automatic Text Substitution adding its results to the Substitution tool.

Here we should add much more afford in my opinion.

t-m-z:
Hm -  I see your points. But these do not touch the users request.
The requirement was to import GADE results - values - into c:geo and/or to import the values into the GC Wizard Formula Editor.

I believe both of your points points are valid, but for a first step not the asked wish of Magnus. GADE at the moment does not provide any formulas but only values.
Thus, 
regarding 1.) the formula editor should be able to export formulas in the c:geo. format. 
regarding 2.) what is the workflow? If someone needs to use the GADE Tool then he/she collects firstly the numbers. Secondly the numbers are to be sorted and thirdly they have to be incorporated into a set of formulas. Maybe it is usefull to add a menu entry "import GADE" which does the sorting of the collected numbers and mapping to variables.

However, for the first step, Magnus was pleased with the solution.

S-Man42:
A user request is not a requirement. Everybody is free to ask for everything, but our task is to think about how to turn it into an app fitting feature. I see, what he was asking for and I think we can offer him something, that will work for him, but also for even more people, not the ones, using Gade.

reg. 2) Press the "Export to Formula Editor" button, which converts the Gade to the tool and opens it directly afterwards. Then he could press "Export Formula Group to cgeo" et voila.

t-m-z:
You are rights 
Sounds valid but another steep learning curve for me

Sent from MailDroid

-----Original Message-----

t-m-z:
done by Mark in a general way

PR #1491, opened 2023-08-24
Api extension

Mike-3:


S-Man42:
aahhh... i was searching for these in the web API... I wondered that I did not find it. But you added this to the server API....

1. Are you planning to add these to the web API for the 3.0.0 start? Would be great!
2. could you please keep the file names lower case. and maybe it makes sense to but the api mappers into a sub directory and try to keep the original tool structure.

PR #1555, opened 2023-12-01
tmz 3.x.x bugfix bowling

t-m-z:
- enable input for 3rd throw in 10th frame
- correct display of scoreboard
- add Hdcp calculation


lib/application/i18n/assets/en.json
```diff
@@ -71,6 +71,8 @@
   "alphabet_name_german2": "German, Version 2",
   "alphabet_name_german2_description": "Ä = AE → 1, 5 | ß = SS → 19, 19",
   "alphabet_name_german3": "German, Version 3",
+  "alphabet_name_danish": "Danish",
+  "alphabet_name_danish_description": "Æ Ø Å → 27 - 29",
```
S-Man42, line 75: Please only. bowlimg bugfix

lib/application/i18n/assets/en.json
```diff
@@ -351,6 +353,7 @@
   "binary_title": "Binary Numbers",
   "binary_description": "Convert binary numbers into decimal numbers",
   "binary_example": "101010 → 42",
+  "binary_length": "Length",
   "binary2image_title": "Binary code to image/ QR Code",
```
S-Man42, line 357: Please only bowling bugfix. And is it necessary to add a new "Length" word? Can you use the units length or maybe there is a common length?

lib/application/i18n/assets/en.json
```diff
@@ -2434,6 +2438,7 @@
   "hexadecimal_title": "Hexadecimal Numbers",
   "hexadecimal_description": "Convert hexadecimal numbers into decimal numbers",
   "hexadecimal_example": "2A → 42",
+  "hexadecimal_length": "Length",
```
S-Man42, line 2441: see above. please try not to add several same new words...

lib/application/i18n/assets/en.json
S-Man42, line 75: is there a source for this count?

S-Man42:
1. Please add ferrrix to testers list
2. I wrote not to add features, only the bowling bugfix... There more I think about it, I believe it is ok in that case, the enhancementa are small enough. However: Nonetheless: I am not even sure if you intended to add them. So if not, pleass keep in mind to check your MRs for their actual inputs...

t-m-z:
I will close this PR. It seems that my "master"-branch is contested

PR #1557, opened 2023-12-08
Bugfix painter (variable length sort)

Mike-3:


S-Man42:
Hi, I wrote a few tests on this issue, that all DO NOT pass. Please have a look at them, for me it is not clear - maybe I don't understand the logic. Maybe we could talk on this.

PR #1593, opened 2024-01-11
Update flutter themes changes

Mike-3:


S-Man42:
For me it's not perfectly resolved. I am missing the base line, which is the divider between the tabs and the lists. And furthermore in my emulator the tap itself looks strange:

It's not a "flat" highlight anymore (which is still in the list elements) but something... ugly... sparkling... bääääh: 
grafik

PR #1595, opened 2024-01-13
Update flutter themes changes (Divider, inactive Icon color)

Mike-3:
296446633-038e549d-baa6-43ed-ae2a-21909371c4c1

Unfortunately I don't know what you mean. To me, the icons look exactly the same in the current web version as they do in the new one.

It is now better ?


S-Man42:
It's not the icons, it's the click effect. Have a look at the pixels in the background of my screenshot! Test it yourself: A click on a list element yield a smooth fade to a brighter background. Now a click on the tab is... ugly... with pixels and something...

Mike-3:
better ?

PR #1609, opened 2024-01-28
add historic conversion table

t-m-z:
important to solve https://www.geocaching.com/geocache/GC32CKH_sakrileg-baldurmorgan-zum-1-


lib/tools/crypto_and_encodings/atbash/logic/atbash.dart
```diff
@@ -1,14 +1,45 @@
 import 'package:gc_wizard/utils/alphabets.dart';
 import 'package:gc_wizard/utils/string_utils.dart';
 
-String atbash(String input) {
+final Map<String, String> HISTORICHEBREW = {
```
Mike-3, line 4: Change to const

Mike-3 (changes requested):

Issue #1554, opened 2023-11-23
UPC Code


S-Man42:
https://www.chegg.com/homework-help/questions-and-answers/universal-product-code-lab-implement-matlab-algorithm-read-upc-universal-product-code-bar--q73770278

S-Man42:
https://coord.info/GCME6T

PR #1626, opened 2024-02-19
UPC-A

S-Man42:
https://github.com/S-Man42/GCWizard/issues/1554


lib/tools/crypto_and_encodings/reverse/api_mapper/reverse.dart
```diff
@@ -1,6 +1,6 @@
 import 'package:gc_wizard/application/webapi/api_mapper.dart';
 import 'package:gc_wizard/tools/crypto_and_encodings/reverse/logic/reverse.dart';
-import 'package:gc_wizard/utils/string_utils.dart';
+import 'package:gc_wizard/utils/string_utils.dart' as strUtils;
```
Mike-3, line 3: Why as strUtils ?

lib/tools/images_and_files/binary2image/widget/binary2image.dart
```diff
@@ -72,7 +102,8 @@ class _Binary2ImageState extends State<Binary2Image> {
   void _createOutput() {
     _outData = null;
     _codeData = null;
-    var image = binary2image(_currentInput, _squareFormat, _invers);
+
+    var image = binary2image(_currentInput!, _squareFormat, _inverse);
```
Mike-3, line 106: null safety ?

lib/tools/images_and_files/binary2image/widget/binary2image.dart
```diff
@@ -16,24 +18,52 @@ import 'package:gc_wizard/utils/ui_dependent_utils/file_widget_utils.dart';
...
+
+  const Binary2Image({Key? key, this.barcodeBinary}) : super(key: key);
 
   @override
   _Binary2ImageState createState() => _Binary2ImageState();
 }
 
 class _Binary2ImageState extends State<Binary2Image> {
-  var _currentInput = '';
+  String? _currentInput;
```
Mike-3, line 30: is this realy necessary ?

S-Man42: yes, because input can come from widget parameter, but not necessarily. So it can be set lately at the first build phase. Do you have another idea how to handle optional input parameters for tools?

lib/tools/images_and_files/binary2image/widget/binary2image.dart
S-Man42: see below

lib/tools/crypto_and_encodings/reverse/api_mapper/reverse.dart
S-Man42: Because there a two reverse functions: in the logic and the strUtils.

PR #1631, opened 2024-03-01
Nonogram

Mike-3:


S-Man42:
1. I reported you a crash with a simple structure.
2. Load file (especially image) does not resize the board. So the loaded image cannot be shown.

Enhancements:
- I personally would limit the size of loadable images... maybe to 50x50 px or something. nobody should create a widget with full-HD or more cells. Maybe we can resize the loaded images on our own?
- I am not quite sure, but on overviewing I saw many different data fields: the rows field (which in fact is the cells field, right?), the rowcounts, the rowHints... All depend on each other. But for example on loading the image, I saw only the setup for the rows data, so the the rowCount for instance is not inconsistent to the data sizes in the rows object. Please check if I am right and correct these data inconsistency if so.
- Add preview for generator

Mike-3:
Add preview for generator -> is icluded

Mike-3:
I am not quite sure, but on overviewing I saw many different data fields: the rows field (which in fact is the cells field, right?), the rowcounts, the rowHints... All depend on each other. But for example on loading the image, I saw only the setup for the rows data, so the the rowCount for instance is not inconsistent to the data sizes in the rows object. Please check if I am right and correct these data inconsistency if so.

I don't know what you mean. The master are the rows. The count of hint rows definied the rows count, The same for the columns.

Mike-3:
I reported you a crash with a simple structure. -> reduce iterations (100 -> 20)

PR #1647, opened 2024-03-25
Word search

Mike-3:
Special option Falling down mode (example GC68ZV6)


lib/application/searchstrings/assets/en.json
```diff
@@ -671,6 +671,7 @@
   "weather_w": "past",
   "weather_ww": "present current forecast",
   "window": "windows",
+  "word_search": "word search letter grid mystery sleuth salad",
```
S-Man42, line 674: wordsearch and lettersearch as one word, if somebody doesn't search with space. 
please leave mystery out, as the entire app is for mysteries and therefore not a good search key. for grids there is a common search key "grid", please use this in the registry instead.

S-Man42:
We're close to the end :)

But... spaces still don't work, right? So, if the letter grid input contains spaces, you ignore them. It's not quite clear to me, why. I can easily imagine a grid which is a normal text with separate words, something like

```
ICH BIN EIN
BERLINER IN
BRANDENBURG
```

Search for:
"IRR" will only work, if the spaces are set correctly.

Maybe somebody then wants do search for "BIN EIN" or "E U" - so suppress spaces seems like a to hard limitation, I think.

--------------

Second, to enhance the UX: Replace "Start" with "Search" and after the first click I would change the text to "Search more" or "Continue search". This makes clear, that you would go for another round and don't start from the beginning.

Maybe it could make sense to put some thoughts into: What if the letter grid will be changed after first search. In my opinion, this should trigger a reset (or better a dialog like: "If you change the letter grid, this will reset the current search. Do you want to continue?").

-------------

Last but not least: Please change the ExpandableDividers into normal dividers. For me it makes no sense to hide a single Textfield, escpecially the search field, which you will often change during the process. To be honest, I even would leave the dividers completely from my point of view and just but a hint text to the TextFields to save horizontal space...

PR #1720, opened 2024-07-01
Feature judoon added

olliklee:
Reworked and splitted from former, deleted pull request


S-Man42:
Please resolve conflicts here.

olliklee:
reverted the last and wrong commit. still conflicts?

olliklee:
done

lib/application/registry.dart
```diff
@@ -1086,6 +1087,19 @@ void initializeRegistry(BuildContext context) {
...
         searchKeys: const ['ipcodes']),
+    GCWTool(tool: const Judoon(), id: 'judoon', categories: const [
+      ToolCategory.CRYPTOGRAPHY
+    ], searchKeys: const [
+      'judoon',
+    ], licenses: [
+      ToolLicenseOnlineArticle(context: context,
+          author: 'Whooligan',
+          title: 'Whooligan\'s Judoon Language Translator',
+          customComment: 'This tool is inspired by the site https://whooligan.weebly.com/index.html'
```
S-Man42, line 1098: I would add this forum as a separate source. So one source is the forum, another the tool. And this way, both sources will be credited with name and even the link.

I would not add the text (Last updated 4 years ago) which would make us update this every year. Either you want to add the last date (which I would not recommend here), then you can use the year/month/date fields or you simply leave it. It is not necessary to put this into the licenses list in my opinion. Maybe you could use the date for a possibly specific post of the forum?

lib/tools/crypto_and_encodings/judoon/widget/judoon.dart
```diff
@@ -0,0 +1,73 @@
...
+            setState(() {
+              _currentInput = text;
+            });
+          },
+        ),
+        GCWText(
+          text: i18n(
+              context,
+              (_currentMode == GCWSwitchPosition.left)
+                  ? 'judoon_unknown_character'
```
S-Man42, line 52: Is this a language specific thing? Otherwise please use the global "UNKNOWN_ELEMENT" instead. We have already several places where we translate into unknown, so this would be consistent (althought not quite perfect yet...). So you don't need the description text. If you want to keep the description text, please stay consistent and make the font size smaller with .copyWith(style: gcwDescriptionTextStyle)

lib/application/registry.dart
```diff
@@ -1086,6 +1087,19 @@ void initializeRegistry(BuildContext context) {
         id: 'ipcodes',
         categories: const [ToolCategory.SCIENCE_AND_TECHNOLOGY],
         searchKeys: const ['ipcodes']),
+    GCWTool(tool: const Judoon(), id: 'judoon', categories: const [
```
S-Man42, line 1090: I would not add it directly under Cryptography but under the LanguageGameSelection where similar things already collected. What do you think? (PigLatin, etc...)

lib/application/i18n/assets/en.json
```diff
@@ -3175,6 +3175,11 @@
   "irrationalnumbers_search_title": "Search Occurrence",
   "irrationalnumbers_error_nonumbers": "Please enter at least one number.",
   "irrationalnumbers_search_description": "Searches all occurences of number groups within the first 1 000 000 digits",
+  "judoon_title": "Judoon",
+  "judoon_description": "Each letter is spelled with a special 'o' loud ; inspired by 'Doctor Who'",
+  "judoon_example": "plo no flo = one",
```
S-Man42, line 3180: Instead of =, for consistency please use the -> character

olliklee: changed

lib/application/registry.dart
olliklee: There is no category like that: PigLatin and similar completely left out the Category list.
I moved it between 'Duckspeak' and 'PigLatin' and removed the category.

lib/tools/crypto_and_encodings/judoon/widget/judoon.dart
olliklee: removed

lib/application/registry.dart
olliklee: all changed and refactored

lib/application/registry.dart
olliklee: ???

olliklee: changed

PR #1742, opened 2024-07-06
fixed code in symbol_table_data.dart

olliklee:
Fixed: The setting for case_sensitive was set to true even if the config_file setting was set to false.


olliklee:
braille and other symbol tables correctly uses the case_sensitive option in their config.files.

PR #1744, opened 2024-07-06
Update symbol_table_data.dart

olliklee:
oops
 != true replaces by == true


olliklee:
hmm in my branch was it correct. github is hard to me

PR #1747, opened 2024-07-08
symbol ice lolly ding

olliklee:
added symbol table

### DIFF
--- a/lib/application/registry.dart
+++ b/lib/application/registry.dart
@@ -474,6 +474,8 @@ import 'package:gc_wizard/tools/wherigo/urwigo_hashbreaker/widget/urwigo_hashbre
 import 'package:gc_wizard/tools/wherigo/urwigo_text_deobfuscation/widget/urwigo_text_deobfuscation.dart';
 import 'package:gc_wizard/tools/wherigo/wherigo_analyze/widget/wherigo_analyze.dart';
 
+part 'package:gc_wizard/application/tools/tool_licenses/specific_tool_licenses.dart';
+
 List<GCWTool> registeredTools = [];
 
 void initializeRegistry(BuildContext context) {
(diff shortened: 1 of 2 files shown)


olliklee:
Could you please check if zip works? I can't zip it with the external tool in lack of Windows. If it works, I could suggest the free tool I used for future Mac developers.

lib/application/i18n/assets/en.json
```diff
@@ -5022,6 +5022,8 @@
   "symboltables_hylian_wind_waker_description": "From the video game 'The Legend of Zelda: The Wind Waker'",
   "symboltables_hymmnos_title": "Hymmnos",
   "symboltables_hymmnos_description": "From the video game 'Ar Tonelico'",
+  "symboltables_ice_lolly_ding_title": "Ice Lolly Ding",
+  "symboltables_ice_lolly_ding_description": "Ice Lolly Ding Font",
```
S-Man42, line 5026: Please remove the description since it has no additional information and only leads to nearly twice the same translations and therefore unnecessary effort for the translators

olliklee: Check and pushed

lib/application/searchstrings/assets/en.json
```diff
@@ -507,6 +507,7 @@
   "symbol_hylian_twilightprincess_wii": "latinalphabet irregular seriffonts smalltriangles smalldots smallcircles reversedhylian mirroredhylian",
   "symbol_hylian_windwaker": "modern japanese angularsigns",
   "symbol_hymmnos": "artistic hairpin bobbypin labyrinth blackdots",
+  "symbol_ice_lolly_ding": "icelollyding",
```
S-Man42, line 510: think properly about search phrases. What about lollies or icelollies. candy? sweets? icecream? such things... sorry mist that in the first review loop

lib/tools/symbol_tables/_common/assets/ice_lolly_ding/config.file
```diff
@@ -0,0 +1,3 @@
+{
```
S-Man42, line 1: i believe this file is not necessary since it is the default...

olliklee: that's true -> deleted

lib/application/searchstrings/assets/en.json
olliklee: Added some. There are more words for the same thing in German for example. Waiting for merge to add that in Crowding

S-Man42:
You have conflicts. This cannot be merged. Please resolve the conflicts.

lib/application/registry.dart
```diff
@@ -4591,7 +4648,20 @@ void initializeRegistry(BuildContext context) {
     GCWSymbolTableTool(symbolKey: 'braille_de', symbolSearchStrings: const [
       'braille',
     ]),
-    GCWSymbolTableTool(symbolKey: 'braille_en', symbolSearchStrings: const [
+    GCWSymbolTableTool(symbolKey: 'braille_en', Ich habe eine Liste der Begriffe erstellt, die ich am häufigsten gesehen habe.
```
S-Man42, line 4651: Are you quite sure, that this compiles? I don't think so. Please DO NOT put commit message into the code!

olliklee: that text accidently slipped into this branch. changed

PR #1754, opened 2024-07-11
Enhancement numeral words. typo. searchstring enhanced

olliklee:
Option to delete spaces between the numbers to make the workflow with coordinates easier and better readable.
Corrected a typo in German translation and added 'zahlenwoerter' to German search strings.


S-Man42:
Correcting translations ONLY in Crowdin!

Otherwise it cause conflicts with the translation platform!

S-Man42:
Could you please give an example for your option? Sometimes spaces make sense, especially if it comes to multi-digit numbers to differ between 10 and 1 and 0

S-Man42:
... and the tradeoff of adding a new widget which makes the tool look more complicated is not worth it... (a long and old topic: how many options you add to make it as flexible as possible vs. make the tool a looking as simple as possible with the tradeoff of having edge cases  not being handled smoothly

S-Man42:
having spaces between numbers is a well known and common problem throughout the entire GCW. We always decided FOR the spaces although knowing about the coordinate issue. However when it comes to the problem, the edit function of the internal copyboard helps you quite well (or the usage of the replacement tool)

olliklee:
I just noticed the editable clipboard after you mentioned it :). I thought the gear symbol was only for changing the settings. Not the content. I see clearer now

S-Man42:
@t-m-z  How do you like another switch for space suppression here? I see both advantages and disadvantages and don't know if i like it :D

@olliklee Pls mind the conflicts here if Thomas is Pro change
@olliklee Pls remove the de.json changes and do it in Crowdin instead

t-m-z:
I am not really convinced. The button does not really solve the problem
image

PR #1755, opened 2024-07-11
Bugfix remove user alphabet

Mike-3:
I can't reproduce the problem in the emulator. Check in the alpha version necessary.


S-Man42:
Did you check your change for backwards compatibility? I guess it won't work for currently saved user's alphabets.

(and I would feel much better to know what the actual problem is before chamging something that deeply...)

Mike-3:
This is no problem. It's only string compare. We create only for new alphabets a longer key.

PR #1756, opened 2024-07-11
Dart 3.0 update

Mike-3:


S-Man42:
why this is necessary to change the switch into if?

Mike-3:
The compiler has a problem with this. This is not more allowed.

S-Man42:
öhm.... ok..... ähm.... this sounds strange

PR #1757, opened 2024-07-11
Change of delete-dialog in favorite list

olliklee:
Improving user experience. Deleting from favorites list gives the user a clearer alert box

### DIFF
--- a/lib/application/i18n/assets/en.json
+++ b/lib/application/i18n/assets/en.json
@@ -951,6 +951,8 @@
   "common_decrypt": "Decrypt",
   "common_deletealtert_title": "Really remove?",
   "common_deletealtert_text": "Do you really want to remove '%s1'?",
+  "common_deletealtert_favorite_title": "Remove favorite?",
+  "common_deletealtert_favorite_text": "Do you really want to remove '%s1' from your favorites list?",
   "common_description": "Description",
   "common_details": "Details",
   "common_distance_hint": "Enter distance",
--- a/lib/common_widgets/dialogs/gcw_delete_alertdialog.dart
+++ b/lib/common_widgets/dialogs/gcw_delete_alertdialog.dart
@@ -6,3 +6,36 @@ void showDeleteAlertDialog(BuildContext context, String deleteableText, void Fun
   showGCWAlertDialog(context, i18n(context, 'common_deletealtert_title'),
       i18n(context, 'common_deletealtert_text', parameters: [deleteableText]), onOKPressed);
 }
+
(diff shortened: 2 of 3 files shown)


S-Man42:
1. DO NOT put any comments into the code whoch actually explains the change. They will be merged and maybe stick to the codebase forever. What you need is a commit comment or a PR comment. I cannot merge anything which such in the code!

2. This is the wrong place for your code. The delete dialog is a general/common widget useable for any widget in the GCW. You favorite dialog instead is only useable for the favorites and should be placed at the favorites widget instead and made private

3. To be honest. I don't like your idea. Of course you could create a specific dialog for any tool, but this means: new translations, uncentralized code,... We try to do the other way round to keep the code as handy as possible. So, creating a special tool dialog just for a new text is not the way. I dont see why "Do you want to delete XYZ?" really is worse than "Do you want to delete favorite XYZ?" when zhe5 tradeoff is additional code and new, nearly equal translations instead a still very flexible code. However, if you really believe, this tradeoff is really a very necessary improvement (and inconsistent to everything else!) then there is a way to use the normal dialog widget with custom text instead of creating a new dialog method.

olliklee:
1. Ok, didn't know that. Comment submit is a good idea. In an extra branch?

olliklee:
Bildschirmfoto 2024-07-23 um 13 22 33

olliklee:
https://github.com/user-attachments/assets/74692318-a1fe-4fbe-af9b-9448422cd461

I would really like to know if it's cool to be asked if you want to delete the whole tool completely when you click on a selected favourite star...
It is a horrible user experience to get this popup in this context. Is it just me who sees this as critical?

t-m-z:
Do we really need a dialog to confirm untapping the favourite star?
It is not a big deal to add the tool again to the list.
If we agrre to this dialog we should ask "Do you really want to remove this tool from the favourite menu?"

S-Man42:
@t-m-z It was a rewuest from Andy innthe very beginning. He meant that one may tap too often accidentially on the star whilrme scrolling. So I added this.
@olliklee I still. understand your concern... ah bit ;) But I am still not sure if it is really such a.big issue to do such a change. And if so, I am still not convinced with the way of the solution. But I didn't had the time to have a deeper look at this because of the many other much more important topics (still have hundreds of symbol licenses open!). Sorry, this exactly is lowest prio at the moment for me

t-m-z:
Okay. I understand.

I am with Olli that the Dialog can he misunderstood.
So we should think about the message.
It is not about deleting a tool but removing it from a list.

Sent from MailDroid

-----Original Message-----

PR #1765, opened 2024-07-15
Postcode

Mike-3:


lib/application/searchstrings/assets/common.json
```diff
@@ -327,6 +327,7 @@
   "pokemon": "pokemon code pikachu",
   "pollux": "pollux",
   "polybios": "polybios polybius transposition",
+  "postcode": "",
```
S-Man42, line 330: what about "postcode", "zielcode", "deutschepost",  "dhl", ...

S-Man42: linear30, ... ? possible searchkeys too?

lib/application/registry.dart
```diff
@@ -1256,10 +1257,17 @@ void initializeRegistry(BuildContext context) {
...
-      'music'
-          'music_notes',
+      'music',
+      'music_notes',
       'piano',
     ]),
+    GCWTool(tool: const Postcode(), id: 'postcode', categories: const [
+      ToolCategory.SCIENCE_AND_TECHNOLOGY
+    ], searchKeys: const [
+      'postcode',
```
S-Man42, line 1267: i'd add "binary" here too

S-Man42: and 'barcodes'

S-Man42:
Hm, indeed, I really like your work, however, I believe you could extend it a bit in the UI.

1. Decrypt: You could simply return the number. like 01234 02 36 2 or something. I personally would find it very interesting to addionally get an explanation of the decryption as the wikipedia does, what do you think about it in a "monospace" textfield?
grafik
2. Encrypt: What if you don't know about the components and simply have a number. I would love to see an opportunity to simply enter a number string. Are there really 0-99 error codes? So is it necessary to have an integer spinner up to 99?

S-Man42:
It makes sense to link the checksum part also into our checksum section.

S-Man42:
Pls don't use test cases from evil kr....de Maybe you could find some from other websites.

Pls add sources to registry

Mike-3:
1. I insert the extension, but we have a size problem. We can not show 80 characters in a line (Smartphone). Landscape format only.

> 2. Encrypt: What if you don't know about the components and simply have a number. I would love to see an opportunity to  simply enter a number string. Are there really 0-99 error codes? So is it necessary to have an integer spinner up to 99?

I think you mean the 'Fee protection code'. I have no information, which numbers are not valid

S-Man42:
1. Valid point. However we have the problem in many cases the smaller the display will get. But maybe one could think about making the CodeTextField horizontally slideable?
2. Check

t-m-z:
Have a look at Befunge, Battleship etc. The CideTextField is horizontally slideable


Sent from MailDroid

-----Original Message-----

Mike-3:
done

Mike-3:
> It makes sense to link the checksum part also into our checksum section.

on todo after merge

Issue #1527, opened 2023-10-03
Wedding anniversaries


Birdy2404:
https://www.welt-der-zahlen.info/hochzeitstage.html
Incl. searchfield like "Porzellan" will show 9 and 20 years
Problem could be the regional differences especially with 6 1/4; 12 1/2; ...
https://de.wikipedia.org/wiki/Hochzeitstag
https://en.wikipedia.org/wiki/Wedding_anniversary
https://es.wikipedia.org/wiki/Aniversario_de_bodas

olliklee:
I'm working on a json file for that. For now are following languages  integrated:  EN/GB, EN/US, DE, NL, SW, DK, IT, ES, FR.
It's not working with crowdin, because of partially completely different representing words for different years.
Any wishes for more languages?
How shall the tool look and which category? Maybe like Zodiac Signs in Miscellaneous Tools?

S-Man42:
I would keep it more simple like IP Schutzklassen

olliklee:
This issue can be closed, i think.

PR #1774, opened 2024-07-17
Wedding anniversaries (#1527)

olliklee:


S-Man42:
grafik
- Germany 54 is missing in the translations
- Please stay convenient with the code basis: Strings usually in ' not in "
- I would definitely add a sort button as discussed ;)
- Due to the copyright discussions: We should start to add the sources (sometimes we already did). So could you please add a wikipedia link top of your data file?

lib/application/searchstrings/assets/en.json
```diff
@@ -688,6 +688,7 @@
   "weather_n": "totalamount cloudcoverage",
   "weather_w": "past",
   "weather_ww": "present current forecast",
+  "wedding_anniversaries": "weddingday jubilee",
```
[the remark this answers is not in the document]
olliklee: All done except sorting.

lib/tools/uncategorized/wedding_anniversaries/logic/wedding_anniversaries.dart
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+String getItems(WeddingCountries country, String year) {
+  var anniversaryList = countryAnniversaries(country);
+  var items = anniversaryList[year];
+  return (items != null) ? items.join(", ") : '';
+}
+
+/* For future use in advanced coding/decoding
```
S-Man42, line 39: Please remove commented code. It will never be touched again ;)

lib/tools/uncategorized/wedding_anniversaries/widget/wedding_anniversaries.dart
```diff
@@ -0,0 +1,115 @@
...
+  Widget _buildOutput() {
+    Map<String, List<String>> _currentMap = countryAnniversaries(_currentCountry);
+    List<List<String>> output = [];
+
+    if (_currentMode == GCWSwitchPosition.left) {
+      // generate output: Element -> Years
+      output.add([i18n(context, "common_element"),
+        i18n(context, "common_year")]);
+
+      var switchedList = _flipKeyValue(_currentMap).entries.map((e) => [
```
S-Man42, line 75: _flipKeyValue is not necessary. Please use the global/common and with unit tests tested switchMapKeyValue()  method instead ;)

lib/tools/uncategorized/wedding_anniversaries/widget/wedding_anniversaries.dart
```diff
@@ -0,0 +1,115 @@
...
+        " " + i18n(context, e.key.toString()), e.value.join(", ")]).toList();
```
S-Man42, line 76: I am not quite sure what the space is for? Is it just for alignment in the table? Then this is the wrong place. Changing output texts just for styling and design is the wrong way. E.g. copying the text is now more complicated, as you will select a space which you as the use don't expect. And it is not consistent to the rest of the app which does not show the indentation.

However we might think about to change the table widget that way that we can add a padding inside or something. But this will be a global (but consistent) change as this should be a separate ticket.

olliklee: ok. looks strange without padding. I'll take it out

PR #1815, opened 2024-07-25
tmz 3.2.0 enhance bacon issue 1622

### LINKED ISSUE #1622: Bacon enhancement
From Mail "thavorianer Tobi":

Moin GCWizard Team,

es wäre toll, wenn ihr noch folgende Ergänzung zu dem Bacon-Code implementieren könntet.

Dieses Verfahren wurden von Sir Francis Bacon (1561-1626) mit dem Hintergrund entwickelt, geheime Nachrichten so zu übermitteln, dass sie nicht als solche erkennbar sein sollten (Steganografie). Das sich ergebene a/b-Muster, eigentlich eine binäre 5-Bit-Kodierung, kann z. B. in Groß/Kleinschreibung oder Anfangsbuchstaben von Wörtern (A-K für 'a', L-Z für 'b') versteckt werden.

Danke, weiter so & viele Grüße

t-m-z:
maybe this is a simple solution for implementing this request
closes #1622


lib/application/i18n/assets/en.json
```diff
@@ -239,6 +239,8 @@
   "bacon_example": "XYZ → BABABBABBABABBB / 101011011010111",
   "bacon_coding": "Coding",
   "bacon_type_full": "Full Alphabet",
+  "bacon_analyze_casesensitive": "Case sensitive: Upper-/Lower Case",
```
S-Man42, line 242: without had a look at what you actually did, i just saw this: AFAIK there is already a common string for case sensitive. Do you mind thisncould be enough?

t-m-z: I'm searching for the right title. A simple "case sensitive" is not enough. Originally Bacon is a kind of steganography cipher.
E.g. the first letters of the words are used to create the code. So an uppercase letter means A, a lower case letter means B. 
Maybe "Bacon Code derived from lower/upper case of every 1st letter" is a better title.

t-m-z: image

S-Man42:
Hm... I am not quite sure if I really like it. For me it feels like a separate tool. Bacon via Steganography (I don't have yet thought about a good title :P ). But to not complicate and confuse with the pretty simple original bacon cipher I guess it might be helpful to add a new tool on this. maybe linked through a BaconSelector. That should make the "Encryption" part also simpler and only users who really know what they do choose this one... What do you think?

PS: "alphabetposition" -> "alphabet position"

S-Man42:
Pls remind the sources :)

t-m-z:
Valid point. However, the Bacon Cipher is twofold.
First you encode your message with the cipher and then you hide the chiffre within a text or picture.
Regarding hiding in a text there are a lot of possibilities:
- using different fonts
- using different font faces like bold, italic, ...
- using upper/lower case
- using ...
Of course all kinds of hiding can't be realized.

I will think about a selection...

Sent from MailDroid

-----Original Message-----

t-m-z:
This will not solve the UI Problem I believe 

Sent from MailDroid

-----Original Message-----

t-m-z:
However I will think about a solution 

Sent from MailDroid

-----Original Message-----

PR #1819, opened 2024-07-30
Substitution: Generate random substitutions

### LINKED ISSUE #1681: Textersetzung: Random encryption

S-Man42:
closes #1681


Mike-3 (approved):

Mike-3:
I just noticed that if I press it multiple times, the block keeps getting appended. Is that intentional?

Mike-3:
If possible, add tests.

S-Man42:
> If possible, add tests.

Hm, hard to add tests on a frontend function and hard to add test on a totally random result... No I believe, here it makes no sense, since it is not lart of the logic but only an enhancement of the input UI

Mike-3:
I didn't even notice it at first because the new entries were outside of my field of vision.

PR #1820, opened 2024-07-30
3.2.0 feature porta chiffre

olliklee:


lib/application/registry.dart
```diff
@@ -996,10 +997,9 @@ void initializeRegistry(BuildContext context) {
               'https://web.archive.org/web/20240715180948/https://xkcd.com/426/'),
       ToolLicenseAPI(
         context: context,
-        author: '@Crox (geohashing.site)',
+        author: 'crox.net (geo.crox.net/djia)',
```
S-Man42, line 1000: you are changing already changed files. Pls don't do the change here

lib/tools/crypto_and_encodings/porta/widget/porta.dart
```diff
@@ -0,0 +1,96 @@
+import 'package:flutter/material.dart';
+import 'package:gc_wizard/common_widgets/outputs/gcw_default_output.dart';
+import 'package:gc_wizard/common_widgets/switches/gcw_twooptions_switch.dart';
+import 'package:gc_wizard/common_widgets/textfields/gcw_textfield.dart';
+import 'package:gc_wizard/tools/crypto_and_encodings/porta/logic/porta.dart';
+
+import '../../../../application/i18n/logic/app_localizations.dart';
```
S-Man42, line 7: import pls with entire absolut path

S-Man42:
I don't know, is that a tool for the MultiDecoder as well?

lib/application/registry.dart
```diff
@@ -5714,31 +5729,20 @@ void initializeRegistry(BuildContext context) {
...
-          stl._toolLicenseOnlineBookAlphabetumAlphabetaCharacteres,
         ]),
     GCWSymbolTableTool(
         symbolKey: 'alphabetum_gothicum',
         symbolSearchStrings: const [
           'symbol_alphabetum_gothicum',
         ],
         licenses: [
-          ToolLicenseOnlineBook(
-            context: context,
```
S-Man42, line 5734: many of the changes....pls be careful!

S-Man42:
Pls add some tests

lib/application/registry.dart
olliklee: Can you reroll the changes on all other than Porta entries? I merged some changes from the upstream, which weren't done and I couldn't compile it with that incomplete entries.

lib/tools/crypto_and_encodings/porta/widget/porta.dart
olliklee: That was an automatic import from the ice. Hiding behind three dots :)

olliklee:
it shows no conflicts anymore

olliklee:
> I don't know, is that a tool for the MultiDecoder as well?

Sure, it has 4 options to try. But it needs a key/password to work.

lib/application/registry.dart
S-Man42: Unfortunately not quite easily. Best thing is: You take the current registry.dart and copy it manually to your branch and re-add your tool.

S-Man42: As long as this is not resolved, I cannot compile and hence review it...

S-Man42:
Please fix the registry issues. Due to the licenses, your registry is too old and yet the PR would remove some entries.

S-Man42:
Still conflicts, e.g. in registry, i18n. Please merge the current 3.2.0 into your branch

S-Man42:
Fixed your conflicts. Will check the feature asap.

olliklee:
Ok. Thanks

lib/tools/crypto_and_encodings/porta/logic/porta.dart
```diff
@@ -0,0 +1,45 @@
+import 'package:gc_wizard/utils/alphabets.dart';
+
+List<String>? _createTable(int version, {bool classic = false}) {
```
S-Man42, line 3: Please don't return unnamed Lists unless other is possible. It is not simple to debug this. You always have to deal with anonymous indexes, have to do a lookup, how many values are meant to be delivered, which one tells me which content, ...

It is better to create a specific return type like:

class PortaCipherReturn {
    String output;
    String headline; // or whatever
}

Now you are able to see exactly, what field to access and you make it type safe and can easily enhance it with any field of any type.

lib/tools/crypto_and_encodings/porta/logic/porta.dart
```diff
@@ -0,0 +1,45 @@
...
+  if (![1, 2].contains(version)) { return null; }
```
S-Man42, line 4: Mainly the same thing for version. Either you only support a specific number of versions, then you should ensure that only these versions can be used. For example, creating an enum PortaChiffreVersions {LEFT_SHIFT, RIGHT_SHIFT}. Moreover, now you can easily read, what value does what and don't have to solve the riddle what is 1 and what is 2 and what if I enter -42.

lib/tools/crypto_and_encodings/porta/logic/porta.dart
```diff
@@ -0,0 +1,45 @@
...
+  String alphabet = alphabet_AZ.keys.join('');
+  if (classic) {alphabet = alphabet.replaceAll(RegExp('[JKUW]'), '');}
+
+  String head = alphabet.substring(0, alphabet.length ~/ 2);
+  String initialLine = alphabet.substring(alphabet.length ~/ 2);
+
+  List<String> table = [head, initialLine];
+  String line = initialLine;
+  for (int i = 0; i < head.length - 1; i++) {
+    (version == 1)
```
S-Man42, line 15: However: Regarding version: I like the idea of having an integer on this nonetheless. I wouldn't call it version anymore but shiftKey or something. But what speaks again to make it more flexible. Yes, I know, most tools only shift one left and one right but it is the same programmatical effort to shift 2 or 3. Just shift the value of your variable? You don't even need to do it by yourself, just use the rotate function:

var usedAlphabet = classic ? _CLASSICALPHABET : _NORMALALPHABET;

var matrix = <String>[];
for (int i =0; i < usedAlphabet.length; i++) {
matrixLine = Rotator(usedAlphabet).rotate(usedAlphabet, shiftKey + i);
matrix.add(matrixLine);
}

lib/tools/crypto_and_encodings/porta/logic/porta.dart
olliklee: The unnamed list is an anonymous helper function for the togglePorta function.
Like in Polybios. Is it really necessary to convert everything to OOP for a simple function?

lib/tools/crypto_and_encodings/porta/logic/porta.dart
olliklee: Changed to enum PortaTableVersion { v1, v2 }

lib/tools/crypto_and_encodings/porta/logic/porta.dart
olliklee: that looks nice. i will overwork that to the rotator. not sure if it's better readable or maintainable though.

but I wouldn't make it more flexible. because it's 'Porta' per definition.  
no need to overpower it, with non native options. this will also followed by more buttons in the widget, which no one really needs and wants.

lib/tools/crypto_and_encodings/porta/logic/porta.dart
S-Man42: Hm... I didnt see that this was private. So... well... I dont like it, but can stay with it...

lib/tools/crypto_and_encodings/porta/logic/porta.dart
S-Man42: Increasing the logic is preparation for future case. Add 1 or two test cases on that.

No specific need for adding a widget though. Just map widget to shiftKey 1 or -1. If someone sometimes needs this, he would only need to change the wodget without touching the logic.

PR #1827, opened 2024-08-05
Tmz 3.2.0 enhancement dtmf issue 1234

t-m-z:
play dtmf tones


pubspec.yaml
```diff
@@ -23,6 +23,10 @@ dependencies:
   file_picker_writable: ^2.0.3
   flutter:
     sdk: flutter
+  flutter_dtmf: ^3.0.2
+  # flutter_dtmf:
```
Mike-3, line 27: remove comments

lib/tools/science_and_technology/dtmf/logic/dtmf.dart
```diff
@@ -1,3 +1,4 @@
+//import 'package:audioplayers/audioplayers.dart';
```
Mike-3, line 1: remove comments

Mike-3 (changes requested):

lib/tools/science_and_technology/dtmf/logic/dtmf.dart
t-m-z: done

pubspec.yaml
t-m-z: done

PR #1840, opened 2024-08-08
3.2.0 Feature Billiard Balls

olliklee:
Billiard balls symbol table added


Mike-3:
The entries for the search are still missing
lib/application/searchstrings/assets/common.json
lib/application/searchstrings/assets/en.json

lib/application/registry.dart
```diff
@@ -5082,6 +5082,17 @@ void initializeRegistry(BuildContext context) {
...
+    GCWSymbolTableTool(symbolKey: 'billiard_balls', symbolSearchStrings: const [
+      'symbol_billiard_balls',
+    ], licenses: [
+      ToolLicenseImage(
+          context: context,
+          author: 'Billiard Gear Vector by Cgvector.com',
+          title: 'Billiard Balls',
+          sourceUrl: 'https://web.archive.org/web/20240524133920/https://www.freevector.com/billiard-balls',
+          licenseType: ToolLicenseType.CCBY4,
+          licenseUseType: ToolLicenseUseType.REPRODUCTION)
```
S-Man42, line 5094: You drawed the images on your own? (Reproduction) Or you took the original ones?

olliklee: The base jpg was from freevector.com. Each ball single copied into a graphic program, changed size, color depth and added some blur.

olliklee:
> The entries for the search are still missing lib/application/searchstrings/assets/common.json

It's in lib/application/searchstrings/assets/en.json because of the difference between the English 'Billiard' and the German 'Billard':

"symbol_billiard_balls": "poolbilliard",

lib/application/registry.dart
S-Man42: Hm... I would say, it's a copy nonetheless. Reproducing is more meant as a completely new image just with using something you looked at. 

Abmalen statt Kopieren sozusagen. Also wenn man ein Bild sieht, das toll ist, man es aber nochmal selbst macht aus irgendwelchen Gründen, nur die Idee übernimmt... ZB wenn die Rechtslage nicht eindeutig ist, ob man etwas kopieren darf

Btw. Light effects, like blurring, shadows, light spots, etc. always cost Kilobytes. I'd prefer a completely simple flat version with a big number because the fancyness doesn't have any additional value. Without having tried it, but I guess the number on the ball could get relatively small when having the symbol size not that big. I personally prefer function over form here. We don't need shiny pictures, better: simple, clear, space savvy.

@Mike-3 what do you think? You already thought much about our symbol sizes...

lib/application/searchstrings/assets/en.json
```diff
@@ -417,6 +417,7 @@
   "symbol_base16": "base hexadecimal lines point triangles",
   "symbol_baudot": "5characters plusandhyphens plusandminuses plussesandminuses plusesandminuses circlesanddots",
   "symbol_bibibinary": "circle lines",
+  "symbol_billiard_balls": "poolbilliard",
```
S-Man42, line 420: i would add "balls" so make it "poolbilliardballs poolballs poolbillardballs"

Which allows searching for

balls
billard
billiard
poolballs
poolbillard
poolbilliard
billardballs
billiardballs
poolbilliardballs
poolbillardballs
...

And maybe something describing when you don't know that this are pool balls. Color, Colour, Circles,.... Maybe there are already separate searchstrings for color and circle, not quite sure. But if so, you can add them to the keyword list directly instead of repeat them in the json

S-Man42: hmmm... maybe poolbillard and poolbilliard is something for the common.json since it is likely to be the same in every language so one does not need to repeat it

lib/application/registry.dart
olliklee: Würden unanimierte gifs auch akzeptiert? Wenn es um Grössenoptimierung geht? PNG ist immer 8 bit RGB, denk ich.

S-Man42: Topic for @Mike-3

olliklee: Ok. I changed to gif89a and it works like a charm. Check if the quality is ok like that.

Mike-3: I have send Olli a newversion 72kB. With png Images. Only the size (150x159) ist not perfect.

PR #1841, opened 2024-08-09
MultiDecoder  Vanity WordsSearch

Mike-3:


S-Man42:
Maybe "and words" is not necessary. I'd think about an entire dictionary translation when reading this. In fact, it's just a nice little addon to the numbers, which can be hidden instead to avoid too much expectations...

lib/tools/crypto_and_encodings/general_codebreakers/multi_decoder/widget/tools/md_tool_vanity_words_search.dart
```diff
@@ -0,0 +1,112 @@
...
+            name: name,
+            internalToolName: MDT_INTERNALNAMES_VANITYWORDSTEXTSEARCH,
+            onDecode: (String input, String key) {
+              var language = _parseStringToEnum(stringNullableTypeCheck(options[MDT_VANITYORDSTEXTSEARCH_OPTION_LANGUAGE], null));
+
+              var detailedOutput = <VanityWordsDecodeOutput>[];
+              detailedOutput = decodeVanityWords(removeAccents(input.toLowerCase()), language);
+
+              String output = '';
+              int ambigous = 0;
```
S-Man42, line 32: typo "ambiguous"

Mike-3:
done

lib/tools/science_and_technology/vanity/vanity_words_search/widget/vanity_words_search.dart
```diff
@@ -148,3 +112,43 @@ class _VanityWordsTextSearchState extends State<VanityWordsTextSearch> {
     );
   }
 }
+
+String buildOutputString(List<VanityWordsDecodeOutput> detailedOutput, BuildContext context) {
```
S-Man42, line 116: private?

Mike-3: No, used from feature and md-tool.

S-Man42: Ok, please rename it to a more specific method, because otherwise it will be shown up in any other tool as method which sounds like a usable one

S-Man42:
> The (DEU) is automatically generated based on the option set. Therefore, different variants are not necessary, you only have to adjust the configuration.

No, you just need to change the key to value: https://github.com/S-Man42/GCWizard/pull/1841/commits/4b3e6715c9dac06a4d78ec1410a91c6f00ecd3e0

PR #1848, opened 2024-08-17
 tmz refactor punchtape

t-m-z:


query
```diff
@@ -0,0 +1 @@
+aehd
```
Mike-3, line 1: Neccesary ?

lib/application/i18n/assets/en.json
```diff
@@ -3995,8 +3995,19 @@
   "punchtape_example": "GCW → [OO .O ] [ OO.O ] [O  .OO]",
   "punchtape_ccitt_title" : "Punch Tape Telegraph (Baudot/Murray/CCITT et al)",
   "punchtape_ccitt_description" : "Electrical Telegraph with 5 Bit Punch Tape invented by Emile Baudot and Donald Murray, normed by CCITT",
-  "punchtape_mode_binary" : "Mirrored\nBit-Order 54321",
-  "punchtape_mode_original" : "Original\nBit-Order 12345",
+  "punchtape_mode_numberonly" : "Only Numbers",
+  "punchtape_mode_lettersandnumbers" : "Letters & Numbers",
+  "punchtape_mode_baudot" : "Baudot",
+  "punchtape_mode_bitorder" : "Bit Order",
+  "punchtape_mode_bitorder_54123" : "54 123",
```
Mike-3, line 4002: Is this right with space ?

lib/application/registry.dart
```diff
@@ -840,16 +840,6 @@ void initializeRegistry(BuildContext context) {
       ],
       licenses: const [],
     ),
-
-    GCWTool(
```
Mike-3, line 844: Delete Exif Reader ?

Mike-3 (changes requested):

lib/application/registry.dart
t-m-z: I found ExifReader twice in the Registry

query
t-m-z: not really - I have no clue where this file came from

lib/application/i18n/assets/en.json
t-m-z: I will change it to Bit Sequence

lib/application/registry.dart
t-m-z: The entry was twice

PR #1874, opened 2024-09-09
gcw openfile iOS enhancement

olliklee:


lib/common_widgets/gcw_openfile.dart
```diff
@@ -198,27 +224,50 @@ class _GCWOpenFileState extends State<GCWOpenFile> {
...
+    switch (_currentMode) {
+      case OpenFileType.FILE:
+        content = _buildOpenFromDevice();
+        break;
+      case OpenFileType.URL:
+        content = _buildOpenFromURL();
+        break;
+      case OpenFileType.IMAGE:
+        content = _buildOpenFromGallery();
+        break;
```
S-Man42, line 238: Isn't there a "missing default" warning? please add default case... Maybe a simple break or something.

lib/common_widgets/gcw_openfile.dart
```diff
@@ -198,27 +224,50 @@ class _GCWOpenFileState extends State<GCWOpenFile> {
...
+    }
 
     return Column(
       children: [
+        Row(mainAxisAlignment: MainAxisAlignment.spaceEvenly, children: [
+          if (Platform.isIOS)
```
S-Man42, line 244: Unfortunately in some rare cases there is a problem with this structure which could not evaluated in the past. So please use the uglier (`Platform.isIOS) ? <Widgets> : Container()` structure instead to avoid unexpected behavior.

pubspec.yaml
```diff
@@ -35,6 +35,7 @@ dependencies:
   http: ^1.2.1
   http_parser: ^4.0.2
   image: 4.1.3
+  image_picker: ^1.0.0
```
S-Man42, line 38: 1.0.0 is correct? Your license links to 1.1.1

lib/common_widgets/gcw_openfile.dart
olliklee: There is an enum with only those 3 cases.

pubspec.yaml
olliklee: Changed yaml to 1.1.1

S-Man42:
grafik
Please put the button bar UNDER (into) the GCWExpandable widget. It's the divider and header of the GCWOpenFile widget. The buttons above them look false. The previous switch was also under the headline.

S-Man42:
grafik
 On Android you now open a (uncommon) gallery open tool as well, not the file picker! Please change back to the file picker for the device button

S-Man42:
The previous view: Switch under the headline:
grafik

S-Man42:
> grafik On Android you now open a (uncommon) gallery open tool as well, not the file picker! Please change back to the file picker for the device button

Hm, I tested with a previous version on the emulator. It looks the same... Maybe an emulator issue on my system. Don't mind atm! We check it with a nightly on a real device!

olliklee:
Hmm. If i checkout to the latest 3.2.0 and test it on pixel 8 pro simulator, i have excactly the same dialog as after my changes. What do you mean?

olliklee:
I haven't changed the _buildOpenFromDevice widget tho. I'm not sure about the changes you mentioned.

PR #1898, opened 2024-09-17
Refactor spelling alphabets

Mike-3:


S-Man42:
Maybe it makes sense to find a way to allow both spellings? I guess there will be many Germans who may creata a mystery with the alpha version (same with whisky vs whiskey?) just for better support

Mike-3:
I changed it from Map to List. Multiple variants can be defined, such as ALPHA and ALFA. The first entry is used for encoding.

PR #1926, opened 2024-09-22
enables automatically hiding of keyboard on mobiles

olliklee:
With clicking somewhere into the body, the focus is taken from the input field.
That needs some testing.
In Enigma is an issue, because the focus is manipulated there. 
Maybe that behaviour of checking if the text is changed can be solved differently?


S-Man42:
looks good from my pov. But i would ask you for a systematic test through all widgets. There may be some which can make any problems. E.g. all where we draw something like grids (sudoku, grid, logical solver, number pyramid, ...) or figures (braille, predator, telegraphs). Or the periodic table, ...

olliklee:
Works with all of those tools, because the focus is only taken from input fields, when clicked into a non reactive part of the body.

PR #1947, opened 2024-10-05
unix time hot fix - please test

olliklee:
please try out


t-m-z:
please close and use my PR
- hot fix does not solve the problem: the Widget has to be initialized - connected to a variable

PR #1949, opened 2024-10-05
Issue 1943 bugfix unixtime

t-m-z:


S-Man42:
even here... 39 files.

lib/tools/science_and_technology/date_and_time/unix_time/logic/unix_time.dart
```diff
@@ -5,20 +5,41 @@ import 'package:gc_wizard/utils/datetime_utils.dart';
...
+          currentDateTime.year,
+          currentDateTime.month,
+          currentDateTime.day,
+          currentDateTime.hour,
+          currentDateTime.minute,
+          currentDateTime.second),
+      loc: currentDateTime,
+    );
   }
+ print(currentDateTime);
```
S-Man42, line 35: Please remove print

test/tools/science_and_technology/date_and_time/unix_timestamp/logic/unix_timestamp_tests.dart
```diff
@@ -25,13 +25,13 @@ void main() {
...
       {'expectedOutput' : DateTime(1970, 1, 16, 2, 11, 57), 'UnixTimeSTamp' : 1300317},
-      {'expectedOutput' : DateTime(1970, 1, 1, 0, 0, 0), 'UnixTimeSTamp' : 0},
+      {'expectedOutput' : DateTime(1970, 1, 1, 1, 0, 0), 'UnixTimeSTamp' : 0},
     ];
 
     for (var elem in _inputsToExpected) {
       test('date: ${elem['jd']}', () {
         var _actual = UnixTimeToDateTime(elem['UnixTimeSTamp'] as int);
-        expect(_actual.GregorianDateTime, elem['expectedOutput']);
+        expect(_actual.loc, elem['expectedOutput']);
```
S-Man42, line 34: in my opinion the test should be on utc. Otherwise you may get some problems when executing the test on another timezone, isn't it?

S-Man42:
grafik

How can a time be the same UTC and GMT

S-Man42:
grafik
Please return Integer.

lib/tools/science_and_technology/date_and_time/epoch_time/unix_time/logic/unix_time.dart
```diff
@@ -0,0 +1,73 @@
...
+      ((jahr - 1) - 1900) ~/ 100 +
+      ((jahr - 1) - 1600) ~/ 400;
+
+  int tage_seit_1970 = (jahr - 1970) * 365 +
+      schaltjahre +
+      tage_seit_jahresanfang[monat - 1] +
+      tag -
+      1;
+
+  if ((monat > 2) && (jahr % 4 == 0 && (jahr % 100 != 0 || jahr % 400 == 0))) {
```
S-Man42, line 45: Are you quite sure that this is really necessary? It is always a bad idea to do same date calc on your own. Couldnt it just be just:

myDate.difference(DateTime.utc(1970,0,0)).inDays?

S-Man42: Or simply .inSeconds (?)

t-m-z:
> grafik
> 
> How can a time be the same UTC and GMT

I have no idea - this is a screenshot from the emulator:
image

lib/tools/science_and_technology/date_and_time/epoch_time/unix_time/logic/unix_time.dart
S-Man42: It looks like you have to convert your time into UTC instead. Or is this already?

You can't compare a localtime like DateTime.now() with UTC. This causes the offset. When I remember correctly, the Widget returns locale, right? If so, you just need a definition what YOUR input should represent, so it makes sense to add a Timezone Dropdown like in MoonRise or something... This needs to. be converted to UTC internally and then you can compare with Date.utc(1970,1,1)

t-m-z:
> grafik Please return Integer.

also interesting here: there is an integer to be seen
image

S-Man42:
Screenshot_20241013_180831_Brave

Yeah, it's as I thought. The widget returns locale, not utc. That could be the main reason of ever, thing. We should think about what we want to do with this. I think, it should return UTC unless no timezone is defined additionally. But this may affect some more tools (which may currently not work correctly?)

S-Man42:
Regarding Integer-Output. Since my Emulator shows the .0 Double: Maybe you could ensure it by something like explicit yourOutput.toInt()?

t-m-z:
I think I found a way

Sent from MailDroid

-----Original Message-----

lib/tools/science_and_technology/date_and_time/epoch_time/epoch_time/widget/epoch_time.dart
```diff
@@ -0,0 +1,203 @@
...
+      }
+    }
+    return GCWDefaultOutput(
+      child: output.Error.startsWith('dates_')
+          ? i18n(context, output.Error)
+          : _currentMode == GCWSwitchPosition.left
+              ? timeStamp
+              : GCWColumnedMultilineOutput(
+                  hasHeader: false,
+                  flexValues: [5, 2, 4],
```
Mike-3, line 133: use const

S-Man42:
Close due to DateTimePicker fixes. More Information via messaging

PR #1961, opened 2024-10-20
categorization of different waypoints on gpx import

olliklee:
initial commit. original or corrected coordinates are standard black, parking coordinates are marked blue, and other waypoints are teal. I used `const Color ...` in gpx_kml_gpx_import.dart. I thought about putting it into constants.dart but that needs more libraries to import there.  Or as enum, but that's too much code for 2 colors, imho
The used colors and the names for the points are discussable. 
I tried successfully the import with Groundspeak, c:geo and Cachly GPX files


olliklee:
opencaching has no special declaration of parking points, except localized names. not sure about the effort to 
 check against any supported language

PR #1979, opened 2024-10-29
hotfix enigma - please test

olliklee:
I left the former version out commented in the code.


S-Man42:
Please do not leave any code commented. No-one ever will delete this anymore. If we needed any old code we could use the Git version history.

PR #1991, opened 2024-11-05
add stack class

Mike-3:
@t-m-z Please check whether the class also works for the script. It should actually be compatible.


S-Man42:
Since this is a utils class, I'd prefer to add a test.

PR #1994, opened 2024-11-10
Parse zipped GPX files into mapview

olliklee:
Enhances gcw_mapview.dart by adding the ability to import zipped GPX files into the map. Some geocaching apps, such as Cachly (iOS), export GPX files as zipped archives. This enhancement allows users to directly import these ZIP files, streamlining the workflow.


olliklee:
Oh. Yes. Sorry

PR #1995, opened 2024-11-10
Parse zipped gpx files

olliklee:
Enhances gcw_mapview.dart by adding the ability to import zipped GPX files into the map. Some geocaching apps, such as Cachly (iOS), export GPX files as zipped archives. This enhancement allows users to directly import these ZIP files, streamlining the workflow.

### DIFF
--- a/lib/tools/coords/_common/logic/gpx_kml_gpx_import.dart
+++ b/lib/tools/coords/_common/logic/gpx_kml_gpx_import.dart
@@ -32,9 +32,8 @@ Future<MapViewDAO?> importCoordinatesFile(GCWFile file) async {
       if (archive.files.isNotEmpty) {
         var file = archive.first;
         file.decompress();
-        var xml =
-            convertBytesToString(Uint8List.fromList(file.content as List<int>));
-        if (file.name.endsWith('.gpx')) {
+        if (getFileExtension(file.name.toLowerCase()) == 'gpx') {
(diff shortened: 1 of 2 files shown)


lib/tools/coords/_common/logic/gpx_kml_gpx_import.dart
```diff
@@ -18,23 +18,37 @@ import 'package:xml/xml.dart';
...
+    case FileType.ZIP:
       InputStream input = InputStream(file.bytes.buffer.asByteData());
       // Decode the Zip file
       final archive = ZipDecoder().decodeBuffer(input);
       if (archive.files.isNotEmpty) {
         var file = archive.first;
         file.decompress();
+        var xml =
+            convertBytesToString(Uint8List.fromList(file.content as List<int>));
+        if (file.name.endsWith('.gpx')) {
```
Mike-3, line 37: use getFileExtension function
use fileExtension(FileType.GPX) 
check lower and upper case (name.toLowerCase())

Mike-3 (changes requested):

lib/tools/coords/_common/logic/gpx_kml_gpx_import.dart
olliklee: if (fileTypeByFilename(file.name.toLowerCase()) == FileType.GPX) {

or this

if (getFileExtension(file.name.toLowerCase()) == 'gpx') {

What is nicer?

PR #2031, opened 2024-12-02
Game of life limit max size

Mike-3:
limited to 200

### DIFF
--- a/lib/tools/games/game_of_life/widget/game_of_life.dart
+++ b/lib/tools/games/game_of_life/widget/game_of_life.dart
@@ -97,7 +97,7 @@ class _GameOfLifeState extends State<GameOfLife> {
         GCWIntegerSpinner(
           title: i18n(context, 'gameoflife_size'),
           min: 2,
-          max: 999,
+          max: 200,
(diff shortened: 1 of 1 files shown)


S-Man42:
I believe the limit is way too low. There are incredible GoL simulations with way more cells. Is there a reason to not set the limit, lets say up to 1000 or something?

Mike-3:
I tested it with 1000 on my smartphone. It wasn't usable. Does it work with 1000 on your phone?

But I have to say that I randomly tested and used 200. Maybe more would work. I just thought that no one would be able to enter/operate it anyway.

t-m-z:
It depends. There are so called RLE-Files to be found in the internet which contain a starting scenario which then develops.
RLE means Run Length Encoded.


Sent from MailDroid

-----Original Message-----

Mike-3:
We could also try to speed up the painter. Instead of 1,000,000 rectangles, only 2,000 lines.

S-Man42:
Yes, maybe a good approach. Paint cells only if they are filled. Otherwise paint lines. Much better

PR #2048, opened 2024-12-24
FormulaSolver: Empty values

### LINKED ISSUE #2022: Formula solver
If there is an empty range variable in the values list, all former calculated correct formulas are getting invalid.

S-Man42:


S-Man42:
@Mike-3 I found another bug which I am not able to fix quickly due to keyvalue widget. It should not be possible to enter a formulavalue key without a letter. You cannot create such, but it is possible to update an already existing - at least in the widget. It is not stored, but the widget is updated nonetheless. It should be prevented that the widget executes the update and should either return to the original value before the update or do not close the edit view. Could you please have a look at this?

PR #2054, opened 2025-01-03
key value editor valid edited key

Mike-3:
change valid edited value

https://github.com/GCWizard/GCWizard/pull/2048#issuecomment-2561740553

wait for branch 3.4.0


Mike-3:
If we need to make a fix, it will be included. But I don't think it's necessary to release a fix specifically for this.

S-Man42:
Hm... for me it was inexpected/unwanted behaviour. But as you like. It's not that urgent

Issue #2060, opened 2025-01-05
rename symbol tables and common_symbols.dart according to the TTF and Unicode specification for glyph names


t-m-z:
e.g.
E_acute => Eacute
e_acute => eacute

As an interim workaround:
add the correct glyph names to common_symbols.dart

Background: 
Using tools to convert a TTF file to a symbol table causes a lot of manual work to adjust the names.
It would ease the work when we stick to the standard.

Birdy2404:
@t-m-z means you add the "wrong" glyph names to the dart as an workaround and we need to extract all glyphs with the wrong name from the zip and zip it again after renaming it?

t-m-z:
No and yes.

We add the correct glyph names to common_symbols.dart to make future work easier. Olli is working on a python script which converts a TTF to a ZIP of PNGs.

Step by Step we check the ZIP-Files and correct the glyph names. The ZIP-Files are not needed to be unzipped.

Mike-3:
If you make a list of the new and old names, I could write a tool (adaptation of the Zip Tool) to update all the names in the zip files.

PR #2072, opened 2025-01-10
initial - preparation for symbol name translations

olliklee:
There is a translation function in the translate_symbols.dart which takes the old filename and returns the new filename.
The translation dictionary there is partially handmade. 
You can use that function for the batch renaming of the zip files.  (#2060)

The  _COMMON_SYMBOLS dictionary has some more information about the characters. This is quite not necessary right now, but keeps some hints for the signs (unicode hex, description) to make it clearer and easier, if we need to add new characters.


S-Man42:
I think your replacement might be a good approach. However, I am curious if the new list really tackles all characters from the old list. Did you check for certain exceptions? Just for curiousity

t-m-z:
Fantastic work.

But please check dash, minus.
Typography knows dash, minus, emdash, endash and hyphen which have different lengths.

olliklee:
The file names should not have those different minus/dash/hyphen stuff. If the glyphnames had those specialities, it should already have been translated in that former name conventions. The translations were made with the unicode characters. So we are safe.

S-Man42:
what about this thing?

PR #2089, opened 2025-01-18
check range to be calculated - closes #2084

### LINKED ISSUE #2084: Mersenne Numbers: Max too high
Currently it is tried to calc 2^111111 as max threshold. Should be corrected to something much smaller :)

t-m-z:
it is a tricky issue because Mark noticed an effect which is determined by
- size of range, hence amount of numbers which have to be calculated
- amount of digits per number

Hence there is no common solution. I have added an alertDialogue to remind the user to check

### DIFF
--- a/lib/tools/science_and_technology/number_sequences/_common/widget/numbersequences_range.dart
+++ b/lib/tools/science_and_technology/number_sequences/_common/widget/numbersequences_range.dart
@@ -5,6 +5,7 @@ import 'package:gc_wizard/application/i18n/logic/app_localizations.dart';
 import 'package:gc_wizard/common_widgets/async_executer/gcw_async_executer.dart';
 import 'package:gc_wizard/common_widgets/async_executer/gcw_async_executer_parameters.dart';
 import 'package:gc_wizard/common_widgets/buttons/gcw_submit_button.dart';
+import 'package:gc_wizard/common_widgets/dialogs/gcw_dialog.dart';
 import 'package:gc_wizard/common_widgets/dividers/gcw_text_divider.dart';
 import 'package:gc_wizard/common_widgets/outputs/gcw_columned_multiline_output.dart';
 import 'package:gc_wizard/common_widgets/outputs/gcw_default_output.dart';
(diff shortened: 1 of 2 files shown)


t-m-z:
After all - it is the colummMultiLine Widget - and not the calculation

lib/tools/science_and_technology/number_sequences/_common/widget/numbersequences_range.dart
```diff
@@ -77,8 +78,15 @@ class _NumberSequenceRangeState extends State<NumberSequenceRange> {
           },
         ),
         GCWSubmitButton(onPressed: () {
+          int countLines = (_currentInputStop - _currentInputStart).abs();
           setState(() {
-            _calculateRange();
+            showGCWAlertDialog(
```
Mike-3, line 83: Does that make sense, even if I only have 10 rows determined, for example?

t-m-z:
Yes. For example 10 rows of Mersenne in the range 111100 - 111110 would cause problems.

Up to now I have no idea how to calculate this.

Sent from MailDroid

-----Original Message-----

Mike-3:
I'm not so happy that you have to press a button every time. Although it does start an executor, which you can cancel.
But I'll merge it.

t-m-z:
The executor can only be stopped while calculating. Depending on the length of the calculated number this could crash.

After the calculation is done the widget will be created.

Depending on the length of the numbers and the amount of the lines the creation of the widget could crash.

 

The crash depends on

*	Kind of number sequence
*	Size of range
*	Starting index of range
*	Size of calculated numbers

 

Up to now I have no idea how to calculate a recommendation about possible crashes.

 

 

Von: Mike-3 ***@***.***> 
Gesendet: Monday, 20 January 2025 17:24
An: GCWizard/GCWizard ***@***.***>
Betreff: Re: [GCWizard/GCWizard] check range to be calculated - closes #2084 (PR #2089)

 

I'm not so happy that you have to press a button every time. Although it does start an executor, which you can cancel.
But I'll merge it.

S-Man42:
@Mike-3 The point is, that the Cancel Button will not help in anyway. The long-lasting action is the widget-creation, not he logic calculation. The cancel button only interrupts the logic execution, not the final widget output. This is why we need a refactoring of the MultiLineDingsensWidget to force it lazy loading, not building all lines at once.

Mike-3:
OK, I understand that it can crash. But the dialog doesn't protect against that. Then we should do a try/catch instead.

There is an issue (#2084) about the problem with the output. I'll take a look at it soon.

PR #2102, opened 2025-01-26
Issue 0000 feature display id3tag

t-m-z:
first idea


S-Man42:
For me it should be part of metadata view, not a separate tool...

t-m-z:
Is a point, but maybe we should start small with two tools:
- Metadata for images, which is already complicated 
- Metadata for sound
And combine these two tools later on

Of course there is also metadata in PDF etc.

S-Man42:
I really like the famous "ExifTool" which is in real a metadata tool for simply everything. We should go the same way. Simply call it Metadata (maybe Exif in description) and add every new Metadata for every new format you like.... Then I don't have to think about what tool I need for my specific format.

Just my 2ct

S-Man42:
And we should always think about writing metadata! So a cacheowner wants to write (or even delete) them

Issue #80, opened 2020-05-02
Show Outputs as QR Codes


S-Man42:
for better sharing

radlerandi:
or JAB code for more Byte ;-) https://jabcode.org/ https://github.com/jabcode/jabcode

S-Man42:
First step was done by integrating the QR output into the Formula Solver Export: #106

Birdy2404:
https://www.geocaching.com/geocache/GC9W9YV_bunter-mystery-2#
Example for Jab-Code out of the checker
https://jabcode.org/scan

S-Man42:
4.0.0 topic?

t-m-z:
Yes

Mike-3:
https://github.com/Mike-3/GCWizard/tree/jab-code
I started it once. But unfortunately there is a bug in it that I haven't been able to find yet.
If anyone wants to continue.

S-Man42:
I am on the QR output topic.

S-Man42:
QR works, but I try following https://stackoverflow.com/questions/79378258/position-subwidget-always-on-top-of-visible-part-of-parent

PR #2112, opened 2025-01-28
Qr output

S-Man42:
Serves #80


olliklee:
I love that option to present the QR Code if possible. For me it's a so-nice-to-have-feature, that I can't wait to see it integrated...
How about hiding that feature behind the standard copy button, and give the copy button a long press?
Even if you have decided to not use long press. It is absolutely common in any operating system: Why not use this to avoid cluttering the UI with new buttons instead of logical enhance the elements we already have?

Advantages:
- The User will not disturbed by that change. It's more a "Have you seen that? That's a new cool feature!", when it gets discovered.
- We just need to change it at one place. GCWButton.dart with one more optional parameter like bool qr_disabled?
- No changing of screenshots in the manual. No rewriting of descriptions. Just an addition in 'New Features' with the next release.

PR #2129, opened 2025-02-03
3.3.2 lazy initial

olliklee:
Very draft but works like a charm

### DIFF
--- a/lib/common_widgets/outputs/gcw_columned_multiline_output_new.dart
+++ b/lib/common_widgets/outputs/gcw_columned_multiline_output_new.dart
@@ -41,8 +41,7 @@ class GCWColumnedMultilineOutputNew extends StatefulWidget {
       _GCWColumnedMultilineOutputNewState();
 }
 
-class _GCWColumnedMultilineOutputNewState
-    extends State<GCWColumnedMultilineOutputNew> {
+class _GCWColumnedMultilineOutputNewState extends State<GCWColumnedMultilineOutputNew> {
   late final int columns = widget.data.isNotEmpty ? widget.data[0].length : 0;
   late int _currentLimit = widget.maxRowLimit;
   late ScrollController _scrollController;
--- a/lib/common_widgets/outputs/gcw_columned_multiline_output_new.dart
+++ b/lib/common_widgets/outputs/gcw_columned_multiline_output_new.dart
@@ -1,0 +1,213 @@
+import 'package:flutter/material.dart';
+import 'package:gc_wizard/application/theme/theme.dart';
+import 'package:gc_wizard/application/theme/theme_colors.dart';
+import 'package:gc_wizard/common_widgets/buttons/gcw_iconbutton.dart';
+import 'package:gc_wizard/common_widgets/clipboard/gcw_clipboard.dart';
+import 'package:gc_wizard/common_widgets/gcw_text.dart';
+import 'package:image/image.dart';
+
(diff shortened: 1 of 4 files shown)


lib/common_widgets/outputs/gcw_columned_multiline_output_new.dart
```diff
@@ -0,0 +1,268 @@
...
+          text: value,
+          style: _text_style,
+          suppressCopyButton: widget.suppressCopyButtons,
+          copyText: copytext);
+    }
+    return GCWText(text: value, style: _text_style);
+  }
+}
+
+//   var rows = _columnedMultiLineOutputRows();
```
Mike-3, line 163: Remove comments

lib/common_widgets/outputs/gcw_columned_multiline_output_new.dart
```diff
@@ -0,0 +1,268 @@
+import 'package:flutter/material.dart';
+import 'package:gc_wizard/application/theme/theme_colors.dart';
+import 'package:gc_wizard/common_widgets/buttons/gcw_button.dart';
+import 'package:gc_wizard/common_widgets/gcw_text.dart';
+
+import 'package:gc_wizard/application/i18n/logic/app_localizations.dart';
+import 'gcw_output_text.dart';
+
+class GCWColumnedMultilineOutputNew extends StatefulWidget {
```
Mike-3, line 9: replace GCWColumnedMultilineOutput
If I see it correctly, it is compatible. And we don't want to have the same thing in there twice

lib/tools/science_and_technology/iata_icao_search/widget/iata_icao_search.dart
```diff
@@ -1,13 +1,14 @@
...
 import 'package:gc_wizard/application/i18n/logic/app_localizations.dart';
 import 'package:gc_wizard/common_widgets/buttons/gcw_button.dart';
-import 'package:gc_wizard/common_widgets/outputs/gcw_columned_multiline_output.dart';
 import 'package:gc_wizard/common_widgets/outputs/gcw_default_output.dart';
 import 'package:gc_wizard/common_widgets/outputs/gcw_output_text.dart';
 import 'package:gc_wizard/common_widgets/switches/gcw_twooptions_switch.dart';
 import 'package:gc_wizard/common_widgets/textfields/gcw_textfield.dart';
 import 'package:gc_wizard/tools/science_and_technology/iata_icao_search/logic/iata_icao.dart';
 
+import '../../../../common_widgets/outputs/gcw_columned_multiline_output_new.dart';
```
Mike-3, line 10: import with package:gc_wizard/...

Mike-3 (changes requested):

Mike-3:
Change target branch to 3.4.0

S-Man42:
Shouldn't it be branch 4.0?

Mike-3:
Mayby it is possible to fix the header.
Or sortable columns (sorting by is then no longer necessary)
Mayby also in the next step.

S-Man42:
(if header is available)

Sorting: Attention, there are basically 3 types of content:
String: Sort alphabetically
Numbers: Sort numerally
Widgets: Sort not available

The hard part could be to detect String vs Number Column, because currently everything is casted into String afaik.

Maybe a different ticket.

Additionally an optional filter could be nice.

S-Man42:
https://github.com/GCWizard/GCWizard/issues/2132

PR #2133, opened 2025-02-04
Entry spinner

Mike-3:


S-Man42:
4.0?

Mike-3:
In my opinion, this is unnecessary because it is not a big change. It looks the same as before, reacts the same as before. And there is still time for testing.

t-m-z (approved):

PR #2165, opened 2025-02-14
correct uppercase operations for ß

### LINKED ISSUE #2157: Kenny's Code: Issues with cipher texts containing ß character
Affects: [Kenny's Code](https://gcwizard.net/#/kenny)
Version: 3.3.2 (316), only web, not reproducible on iOS

The output gets messed up if the cipher text contains the ß character. I can reproduce this simply by typing in `ß` which gets decrypted to `ßundefined`

Longer example: `ß Fmpmfpmfffmm mfffmm mmm fmpmppfmmfmp.`
Output: `ß Fthismismaftestundefined`
Expected: `ß This is a test.`

t-m-z:
it seems that the uppercase handling of ß causes the issue.
Dart converts ß to SS while the capital ⟨ẞ⟩ was encoded by Unicode in 2008 at (U+1E9E ẞ LATIN CAPITAL LETTER SHARP S).

Background
In 2017, the [Council for German Orthography](https://en.wikipedia.org/wiki/Council_for_German_Orthography) officially adopted a capital, ⟨ẞ⟩, as an acceptable variant in German orthography, ending a long orthographic debate.[[4]](https://en.wikipedia.org/wiki/%C3%9F#cite_note-Long_debate-5) Since 2024 the capital ⟨ẞ⟩ (ligature) has been preferred over ⟨SS⟩ (two letters).[[5]](https://en.wikipedia.org/wiki/%C3%9F#cite_note-Amtliches_Regelwerk-6)

### DIFF
--- a/lib/tools/crypto_and_encodings/kenny/logic/kenny.dart
+++ b/lib/tools/crypto_and_encodings/kenny/logic/kenny.dart
@@ -89,6 +89,7 @@ String decryptKenny(String input, List<String>? replaceCharacters, bool caseSens
   }
 
   // substitute key characters/string with non-writeable characters
+
   var _input = substitution(input, substitutions, caseSensitive: false);
   int chunkStart = 0; // start position of chunk in the original text
   int chunkOffset = 0; // position in the chunk
@@ -148,7 +149,13 @@ String decryptKenny(String input, List<String>? replaceCharacters, bool caseSens
   }
   // restore unused chunks to the original text
   output = _restoreChunks(output + _input, input, 0, substitutionsSwitched);
-  if (!caseSensitive) output = output.toUpperCase();
+  if (!caseSensitive) {
+    if (output == 'ß') {
+      output = '\u1E9E';
+    } else {
+      output = output.replaceAll('ß', '\u1e9e').toUpperCase();
+    }
+  }
 
   return output;
 }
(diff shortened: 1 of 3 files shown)


S-Man42:
Did you ran ALL tests! I believe when changing substitution VERY much will fail. It's a very central function, touching this has a huge impact and should not be done in a simple go-by! I call attention and warning! 

Additionally add ß test for both!

S-Man42:
I would also like to discuss this fix in a bigger round! I am nit quite sure if this is really wanted. Yes, there is an uppercase ß. But many many many algorthms do not care for it and the ss conversion is not only expected but maybe mandatory. I would handle this either with a flag in substitution or - since it seems to yield problem extremely rarely - direct in the kenny code by searching the input of the function and handle it directly there without touching global utils.

I have a big stomach pain on this one!

Nice find nonetheless!

t-m-z:
> Did you ran ALL tests! I believe when changing substitution VERY much will fail. It's a very central function, touching this has a huge impact and should not be done in a simple go-by! I call attention and warning!
> 
> Additionally add ß test for both!

I ran only substitution_test and string_utils_test which ran fine

t-m-z:
> I would also like to discuss this fix in a bigger round! I am nit quite sure if this is really wanted. Yes, there is an uppercase ß. But many many many algorthms do not care for it and the ss conversion is not only expected but maybe mandatory. I would handle this either with a flag in substitution or - since it seems to yield problem extremely rarely - direct in the kenny code by searching the input of the function and handle it directly there without touching global utils.
> 
> I have a big stomach pain on this one!
> 
> Nice find nonetheless!

I understand your concerns toucing an old old and widely used tool. That's why I asked your review.
The issue is raised because of Dart using a deprecated conversion as standard. As soon as Dart changes we have to adopt and find a solution.

An easy/elegant workaround within Kenny is not suitable I believe because of several checks regarding CaseSensitiveness.
If a SS-conversion is mandatory while solving a mystery, the user still can substitute ß to SS manually so I would like to stick to the decision of the german language authority.
Otherwise I would prefer a flag setting the deprecated standard as default.

PR #2168, opened 2025-02-15
re-arrange search bar

### LINKED ISSUE #2166: Search overwrites tabs
Originally I thought that adding a search result to the favourites would have flooded my favourites list with all search results.

However, it only looks like the tabs are still working when search results are displayed, but in reality the tabs are not active and everything is overwritten by the search results.

I think it would be better if the tabs were completely hidden and not active as long as the search is active.

Here is a video that hopefully shows what I mean: https://1drv.ms/v/s!Au3aARolo768lLAqkTHicolRsuKjsQ?e=4T2l5G
(Don´t know why I can´t attach this video here)

t-m-z:
closes #2166

### DIFF
--- a/lib/application/category_views/all_tools_view.dart
+++ b/lib/application/category_views/all_tools_view.dart
@@ -423,37 +423,48 @@ class _MainViewState extends State<MainView> {
 
     var toolList = (_isSearching && _searchText.isNotEmpty) ? _getSearchedList() : null;
 
-    return DefaultTabController(
-      length: 3,
-      initialIndex: Prefs.getBool(PREFERENCE_TABS_USE_DEFAULT_TAB)
-          ? Prefs.getInt(PREFERENCE_TABS_DEFAULT_TAB)
-          : Prefs.getInt(PREFERENCE_TABS_LAST_VIEWED_TAB),
-      child: Scaffold(
-        key: _scaffoldKey,
-        appBar: AppBar(
-            bottom: TabBar(
-              onTap: (value) {
-                Prefs.setInt(PREFERENCE_TABS_LAST_VIEWED_TAB, value);
-              },
-              tabs: const [
-                Tab(icon: Icon(Icons.category)),
-                Tab(icon: Icon(Icons.list)),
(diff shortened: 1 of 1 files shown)


S-Man42:
Should be a 4.0 feature, right?

t-m-z:
Because it's not really a breaking change I am not quite sure to wait for 4.0.0


-----Original Message-----

S-Man42:
Just my opinion: We discussed a 4.0 release with ui changes. That's what I have in my mind. If we decide to not collect all things that we change and release them sequentially, it's ok for me. But thrn I could release my QR button as well or whatever. I understood we want to discuss the 4.0 together, how we want to do this or that and decide the solution and then put all UI changes in one big 4.0 release. That includes a search bar change.

However, that's what I thought. I would like to have a decisson on how you planned this. And please let us make a session for 4.0 or not 4.0

t-m-z:
Valid thoughts. Talking about 4.0.0 with an enhanced UI as a braking change is of course necessary. The question is, which changes are braking or not.

I did not recognizes this search bar behavior for years. I believe I am not alone. So I think this is not a braking change beacause no one will recognize it.
Your QR Button is slightly more. Up to now we have only a copy button. Now we have a QR code button and probably a Save Button would be questionable - so what ab out a Hamburger Menu for Copy, QR and Save?

S-Man42:
Yes but also said that every UI change lead to new screenshots, breaking or not. You said, you want to collect all UI changes to just need the screenshot making once.

But as said, you and @Mike-3  should decide and communicate it. :)

olliklee:
I'm sure only a handful of our users would even notice the change.
I strongly disagree that we should wait until version 4.0 to integrate this fix. In my opinion, it is a FIX, not a UI change, so updating the handbook shouldn't be necessary.

The other thing mentioned: [see my comment PR #2112](https://github.com/GCWizard/GCWizard/pull/2112#issuecomment-2662830395)

Mike-3:
I also think that this is not such a big change that we have to wait until 4.0

Issue #2167, opened 2025-02-15
Map View Open File


hermannslaender:
May I ask why the Open File dialog in Map View is so large?

![Image](https://github.com/user-attachments/assets/f84e7d6a-95c3-45b7-91dd-830be5d1ab9f)

 It´s nearly full screen height.
Is there something I should see but can´t?

olliklee:
Might belong to #1882

S-Man42:
Maybe it makes sense to make this change more globally to all gcw dialogs, not only the open file, because all have a similar problem. #1882

PR #2173, opened 2025-02-18
fixes size of open file dialog

olliklee:
Better size and information about possible filetypes.


olliklee:
Closes #2167

S-Man42:
Maybe it makes sense to make this change more globally to all gcw dialogs, not only the open file, because all have a similar problem. https://github.com/GCWizard/GCWizard/issues/1882

PR #2182, opened 2025-02-21
splitting names and code in about.dart

olliklee:
Refactored for easier maintaining. Removed a lot of double code lines.


olliklee:
Came just in my mind: Are those list variables better declared as anonymous?

PR #2195, opened 2025-03-01
move formula solver logic

Mike-3:
Logic from widget to logic moved
logic in varaible coords used

### DIFF
--- a/lib/tools/coords/variable_coordinate/logic/variable_latlon.dart
+++ b/lib/tools/coords/variable_coordinate/logic/variable_latlon.dart
@@ -75,13 +75,14 @@ VariableCoordinateResults parseVariableLatLon(String coordinate, Map<String, Str
     }
   }
 
-  var calculated = FormulaParser(unlimitedExpanded: true).parse(textToExpand,
-      substitutions.entries.map((e) => FormulaValue(e.key, e.value, type: FormulaValueType.INTERPOLATED)).toList());
+
+  var values = substitutions.entries.map((e) => FormulaValue(e.key, e.value, type: FormulaValueType.INTERPOLATED)).toList();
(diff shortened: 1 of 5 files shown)


S-Man42:
It's not quite easy to me to follow the steps here. The work which needed to be done was: Moving the logic for formula references from frontend to backend. I can see, that you did something like this. But I cannot evaluate, that this was done correctly. Since the logic was enhanced, you should add several test cases to proof the functionality for formula refs.

Mike-3:
I can see if the tests can be added. But no new function has been added. It has just been moved.

S-Man42:
If it still works than it's fine. However tests are mandatory here nonetheless they must ensure to keep the complicated logic, at least to give a complete test harnish for re-implementation.

S-Man42:
{A40C89F3-83B0-4689-8111-8322B79611A7}

I confess, I had an older status, but now, they have different colors for tracks and paths and footways. It was not long ago a mess. So, yes, I confess, it is not bad anymore :)

However, since it is another mapstyle, one should not have one visualization in one language and a different map style in another one. So, I would prefer, adding a layer dropdown for different styles. Maybe a v4 feature?

S-Man42:
When using the German tiles, we need to attribute them correctly. Data still osm.org, but map tiles come from osm.de/FOSSGIS
https://www.fossgis.de/arbeitsgruppen/osm-server/nutzungsbedingungen/

Wish list for layers:
- Transport map (there are different ones, one on the osm.org and another one on the osm.de page)
- Cycle map (also different ones available)
- OpenTopoMap (I love it!)
- OpenSeaMap (not quite sure if possible)
- ...

Always ensure that it is allowed to integrate the maps for our usage purposes!

PR #2198, opened 2025-03-01
Pokemon Unown

Birdy2404:
Added names according to https://www.pokewiki.de/Icognito


Birdy2404:
Version 3.3 instead of 3.4

PR #2202, opened 2025-03-04
handling line breaks

### LINKED ISSUE #2201: ave maria code loses letters on line breaks
If a carriage return is in the encoded text, the last letter for the former paragraph and first letter of the next paragraph disappears.

olliklee:
in decoding: line breaks will replaced with 2 spaces before splitting.
Closes #2201


Mike-3:
add a test

PR #2203, opened 2025-03-04
Formula solver web json import

Mike-3:
Create weblink with parameter extension


lib/tools/formula_solver/widget/formula_solver_formulagroups.dart
```diff
@@ -53,8 +54,32 @@ part 'package:gc_wizard/tools/formula_solver/widget/formula_solver_values.dart';
...
+      }
+    },
+    "parameters" : [
+      {
+        "in": "query",
+        "name": "input",
+        "required": true,
+        "description": "Formula data in json format",
+        "schema": {
+          "type": "string"
```
S-Man42, line 73: this is an openAPI description, which is made for describing JSON objects in endpoints. So I would prefer to have an JSON object description here about how the actual formula object really looks like - instead of saying it's a string. It's JSON and it can be used here to describe it.

Mike-3: better ?

PR #2206, opened 2025-03-10
extended city names on 'de'-language

olliklee:


S-Man42:
I personally don't like the osm.de layout, especially not for hiking. It's not just changing the geographic names, it changes the entire layout and colors. So, I personally vote for either a new layer (and therefore a new widget instead of the toggle button) or not doing it. I would not be happy with that :'(

S-Man42:
{086E81A3-87CE-4876-B354-A119DF33B037}


{A40C89F3-83B0-4689-8111-8322B79611A7}

I confess, I had an older status, but now, they have different colors for tracks and paths and footways. It was not long ago a mess. So, yes, I confess, it is not bad anymore :)

However, since it is another mapstyle, one should not have one visualization in one language and a different map style in another one. So, I would prefer, adding a layer dropdown for different styles. Maybe a v4 feature?

S-Man42:
When using the German tiles, we need to attribute them correctly. Data still osm.org, but map tiles come from osm.de/FOSSGIS
https://www.fossgis.de/arbeitsgruppen/osm-server/nutzungsbedingungen/

Wish list for layers:
- Transport map (there are different ones, one on the osm.org and another one on the osm.de page)
- Cycle map (also different ones available)
- OpenTopoMap (I love it!)
- OpenSeaMap (not quite sure if possible)
- ...

Always ensure that it is allowed to integrate the maps for our usage purposes!

olliklee:
Ok. I understand the rules from fossgis. How does cgeo handle that? I estimate that they have 10,000 times more network traffic than we do.

olliklee:
Just an idea. How about offer own map providers in the settings with the use of own api keys. Configurable in the settings dialog?

S-Man42:
Maybe a list of possible layers with tile download cache to avoid double download. Adding an expiring duration may be useful and additionally (as Locus does) a limitation of max tile downloads per layer and day. So a small map manager is required. But if it works, then it could work with adding even small servers. 

Created issue #2208

olliklee:
I'll take this out for now. This needs more deep thinking about the complete map management and visualization.

PR #2218, opened 2025-03-15
no ad in gold

olliklee:
easier way. determine app name once at start.


Mike-3:
I prefer the other option. Why does it have to save the app name and then reload it again and again?

olliklee:
Only once per start, not again and again. And if someone decides to change the name in expert setting mode, it is resetted on the next start of the app again. But I changed the other version to keep the count of checks minimized. So this PR is obsolete.

PR #2252, opened 2025-04-07
Fakoo segmentdisplay

Mike-3:


lib/application/registry.dart
t-m-z, line 1: Please add license according to https://fakoo.de/fakoo.html

t-m-z:
we should think about renaming the selection and tools because we do not show only Braille but also Fakoo.
If someone does not use the search tool he/she will probably has problems finding Fakoo.

t-m-z (approved):

PR #2268, opened 2025-05-07
Copy unconvertible input in format converter

olliklee:
Please test on other formats.


olliklee:
This is doing exactly what Mark mentioned. Just copies the input without translations.

PR #2339, opened 2025-08-18
Issue 0000 bugfix calendar 3.5.0

t-m-z:
The DatePicker did not get correct Hebrew Dates because of 13 months in the Jewish calendar.
Hence a custom picker had to be created.


Mike-3 (commented):
some calendar tests fail

Issue #2362, opened 2025-11-02
enhancement scrabble, bww


t-m-z:
while solving https://www.geocaching.com/geocache/GC47NVV I had to convert the sentence "go up sand find point note ftf   one log a tub is found" into the word values calculated using scrabble values.

currently this sentence is decoded only letter by letter.

suggestion:
rebuild the ui in a way, that the output will show the scrabble values for each word and also the details for for each word letter by letter

t-m-z:
this should be done in general way that also bww will be improved

t-m-z:
Suggestion for this method:
- input:
   + String sentence
   + function decode eg. scrabble, asciivalue,...
- output
   + List of widgets:
      - Text Widget with list of integer
      - ExpandableTextDevider for details
         + List of ExpandableTextDivider with details per word

S-Man42:
Well: scrabble, ascii, bww, etc. is absolut the same calculation. Only differenf alphabet value maps. So no need for a specific function. The code is quite easy here:

input.split(whatever word delimiter is).map((string word) => letterValue(alphabetMap)).join().

The hard part is to put it into a useable, understandable UI. I suggest a new tool: Letter Values for Words. Ot looks like the original one, but withe several values per calculation.

However in that case it could be hard to find the bww for the 14th goven word. Maybe one shouls think about an enumeration. But maybe that would be overkill.

S-Man42:
I work on this. It's not quite as easy as thought...

t-m-z:
running in the emulator, the switch to differ between letters and words does not do anything

<img width="978" height="1213" alt="Image" src="https://github.com/user-attachments/assets/777f05db-10ff-443c-aff2-8b35c9056655" />

t-m-z:
I have to work on beatnik due to an exception:

The following _TypeError was thrown building Beatnik(dirty, dependencies: [_LocalizationsScope-[GlobalKey#bb484]], state: _BeatnikState#186f8):
TypeError: Instance of 'WhereIterable<int?>': type 'WhereIterable<int?>' is not a subtype of type 'List<int>'

The relevant error-causing widget was: 
  Beatnik Beatnik:file:///C:/Programmierung/GCWizard/lib/application/registry.dart:3793:25
When the exception was thrown, this was the stack: 
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 274:3                                                       throw_
[...]

PR #2368, opened 2025-11-10
#2362 BWW and Scrabble are able to calculate per word group

S-Man42:
@t-m-z Please check Beatnik. BC of its Scrabble connection, I needed to do a small change in the logic code.


S-Man42:
grafik

lib/application/i18n/assets/en.json
t-m-z, line 1: please check grammar of "crosssum_range_iterated_frequency_description"

S-Man42: Is ok IMO. I learned the term "reduced cross sum" in london a few weeks ago. I believe it is more common then "Iterated", which was always a 1:1 translation of myself...

Or what do you mean?

lib/tools/crypto_and_encodings/esoteric_programming_languages/beatnik_language/logic/beatnik_language.dart
t-m-z, line 1: this produces an exception. The workaround ist
for (int i = 0; i < program.length; i++) {
    if (!normalized) {
      value =0;
      List<int?> scrabbleValues = scrabbleTextToLetterValues(program[i], ScrabbleVersion);
      for (int? digit in scrabbleValues){
        if (digit != null) {
          value = value + digit;
        }
      }
    } else {
      value = int.tryParse(program[i]) ?? 0;
    }
    if (value > 0) {
      assembler.add(value);
      scrabbleProgram.add(_formatNumber(value) + ' ' + program[i]);
    }
  }

S-Man42: Ah. Ok. Afterwards I learned about ist scrabbleValues = list.whereType<int>();

very handy and should solve the problem by only filtering the int values, not the int? values. I used it somewhere else. Could.you please check with that approach? This would avoid the additional loop

lib/application/i18n/assets/en.json
t-m-z: it is the sentence of the description: Calculates which reduced cross sums and how often they occur within a range.

I would replace WHICH with THE

S-Man42: ah. Didnt see that.

lib/tools/crypto_and_encodings/esoteric_programming_languages/beatnik_language/logic/beatnik_language.dart
t-m-z:
```
for (int i = 0; i < program.length; i++) {
    if (!normalized) {
      value = scrabbleTextToLetterValues(program[i], ScrabbleVersion)
          .whereType<int>()
          .toList()
          .fold(0, (previousValue, element) => previousValue + element);
    } else {
      value = int.tryParse(program[i]) ?? 0;
    }
    if (value > 0) {
      assembler.add(value);
      scrabbleProgram.add(_formatNumber(value) + ' ' + program[i]);
    }
  }
```
should be the new code

S-Man42:
Check-in

Mike-3 (approved):

PR #2376, opened 2025-12-01
Animated image encoder

### LINKED ISSUE #2302: AnimGif Creator
Add Images
Add Duration per Image or Total Duration (gleichverteilte Zeit)
Option for Loop/Reverse
Possibility to reorder images if opened in wrong order

Create

Mike-3:


lib/application/i18n/assets/en.json
t-m-z, line 1: May rename High to Das/Dot and Low to Silence

lib/tools/images_and_files/animated_image_morse_code/logic/animated_image_morse_code_encode.dart
t-m-z, line 1: code seems ok.

However:
- check use of images of different sizes. Either they should be scaled to the same size or an error should be raised.
- check encoding/decoding. Encoded GIFs aren't decoded properly

t-m-z (changes requested):
see comments

lib/application/i18n/assets/en.json
Mike-3: I'm not sure if many people know the internal designations for Morse code signals (short dot/dit or long dash/dah).

Maybe just on and off?

t-m-z: I believe that cache owner are familiar with punkt/strich/pause oder dash/dot/break/silence etc.

Mike-3:
ready to review. 
Add scale (Mark wish).

lib/tools/images_and_files/animated_image/widget/animated_image.dart
t-m-z, line 1: scale does not work properly
only las loaded filename is shown ind encrypt mode
when switching to decrypt mode the filename is still shown

Mike-3: What's not working correctly with the scaling? Each image is being scaled to the desired scale.

t-m-z:
I did not noticed a way to scale and I would have expected that all images are scaled to the same size. 

Am 16. Dezember 2025 11:01:39 MEZ schrieb Mike-3 ***@***.***>:

Mike-3:
Screenshot_20251216_181139
Scale is under options

Mike-3:
I had also considered scaling, but the images might then be distorted. Therefore, they are all currently scaled by the same factor. Users can adjust them beforehand if needed.

t-m-z (approved):

PR #2398, opened 2026-01-09
3.6.1 piano rework

### LINKED ISSUE #2393: bug in piano frequencies
according to 
- https://sengpielaudio.com/Rechner-notennamen.htm 
- https://inspiredacoustics.com/en/MIDI_note_numbers_and_center_frequencies

the MIDI number is an Integer digit.

We should also consider that there are
- 127 MIDI keys
- 88 Piano keys
- 61 keyboard keys

Hence the whole map has to be adjusted

olliklee:
Changed 1000+ lines of hardcoded map to calculated values.
Added Midi numbers and options.
Changed latin notes to a more common appearance.


S-Man42:
Did you check with other sources? The original ones came from Wikipedia (en or de, I do not know anymore), afaik. So the changed ones should be plauskble against them or there should be a prove, that they are wrong otherwise

S-Man42:
So.please be really careful and sure about these changes. It was a hard piece of work once. I will not check the changes but I'd like to ask you for a double-check tobbe really sure. because once I did massive researches on my own... Maybe I was completely wrong, but maybe not...

olliklee:
I checked against your work.

The latin names has some issues. 
Did you mean Italian instead of Latin?
This is for key 53:
German = cis2/des2
Scientific = C♯5/D♭5
Italian = Do diesis 4/Re bemolle 4 
or shorter Do#4/Re♭4

Where is the source for 
Latin = Do4#?

S-Man42:
To.be honest, no idea :D

Keep your work if you are sure. Don't forget to add your sources :)

Mike-3:
Please change the target branch to 3.7.0. When I this do, I have more changed files.

PR #2405, opened 2026-01-22
scout chiffer

victorhaggqvist:
Scoutchiffer, see eg. https://skultuna.scout.se/information/chiffer/ (in swedish). Not too sure how originated in Sweden it is, but this never the less includes ÅÄÖ. But, opted to not name it so, as eg. not like `Kabouter ABC` described "ABC from the boy scouts netherlands".

Key
image


t-m-z:
Hi Victor, 
Thanks for your great contribution. 
We really appreciate your effort. 

Unfortunately we ha already such an option implemented - the Polybios Cipher. 

This tool offers already an option to build custom polybios squares where you can easily change the headline row and front column. 
Comparing to your PR I see one option: to enhance the current polybios tool to be case sensitive for the headline row and front column.

Screenshot_20260122-230142


Sorry

Knd regards
Thomas

lib/application/i18n/assets/en.json
t-m-z, line 1: I understand using encode/decode - but for consistency reasons I recommend to stick to encrypt/decrypt like the other tools

lib/application/registry.dart
t-m-z: Regarding the GCWTool: we try to add to every tool a license information which shows were we found the tool or the sources we used.
If these sources ar an internet link, we add this website to the internet archive and use that hyperlink.
Maybe the swedish website should be kept as source.

lib/application/i18n/assets/en.json
S-Man42: There are many tools that use encode instead of encrypt, if I remember correctly...

PR #2449, opened 2026-03-16
Punycode

### LINKED ISSUE #2436: Punycode
https://de.wikipedia.org/wiki/Punycode?wprov=sfla1

Mike-3:


lib/application/searchstrings/assets/common.json
t-m-z, line 1: please add RFC 3492 idna internationalizingdomainnamesinapplications

lib/application/registry.dart
t-m-z, line 1: please add  https://datatracker.ietf.org/doc/html/rfc3492 and perhaps the link to the C reference implementation

link to license https://www.rfc-editor.org/rfc/rfc3492.html#page-35

PR #2454, opened 2026-04-01
Ortho by two points

S-Man42:


Mike-3:
The same icon for both ?

PR #2471, opened 2026-04-26
add german bingo calls

t-m-z:
there are no specific german bingo call to be found

found a complete translation at https://gamblingngo.com/de/guides/bingo-calls/

either we add the translation or we keep only the original british ones and think about crowding which means 200 additional strings


S-Man42:
i would just remove the language dropdown.

S-Man42:
even the description... means 49 (?) additional strings, which in fact nobody read...

